PR #9, opened 2021-04-07
[Bug][Windows] Modify the timeout value for Windows installation

Tomorrow9:
Signed-off-by: dianew <dianew@vmware.com>

In some slow environment, the Windows OS installation process takes longer than 600s to get the IP address, so here modify this value to 1800s to fit some specific environment or situations.


keirazhang (approved):
LGTM. Thanks!

PR #59, opened 2021-06-10
Remove spaces in yum repo file

keirazhang:
Photon's yum repo file doesn't accept spaces before and after '=', so remove those spaces. The others os distributions which using yum repo accept repo settings with or without spaces before and after '='.


Tomorrow9 (approved):
LGTM, thanks.

PR #85, opened 2021-07-01
Module nvme is builtin in Photon OS. Do not reload it

keirazhang:
Photon OS doesn't support NVMe hot add/remove on existing NVMe controller. So nvme_vhba_device_ops would fail with Photon OS. This fix will not reload NVMe module in Photon OS as it's builtin.


Tomorrow9 (approved):
LGTM, thanks.

PR #255, opened 2022-03-04
Fix ubuntu_version failure when ISO name has no version

keirazhang:
If the Ubuntu ISO name doesn't contain any version information, like jammy-live-server-amd64.iso, set fact of ubuntu_install_method would fail as ubuntu_version is empty. And the task would fail with below message:

```
2022-03-03 03:53:27,003 | TASK [Set Ubuntu autoinstall method] ***********************
task path: /home/worker/workspace/Ansible_Cycle_Ubuntu_20.04.4_ISO/ansible-vsphere-gos-validation/linux/deploy_vm/ubuntu/prepare_ubuntu_iso_install.yml:21
fatal: [localhost]: FAILED! => {
    "msg": "Input version value cannot be empty"
}
```


Tomorrow9 (approved):
LGTM, thanks!

PR #304, opened 2022-05-16
Remove xorriso from requirement.txt and add to README

keirazhang:
`xorriso` is not a python package but linux tool, so remove it from requirements.txt


Tomorrow9 (approved):
LGTM, thanks!

PR #316, opened 2022-05-26
[Windows] Change to use pvscsi driver in VMware tools 11.3.5

Tomorrow9:
Signed-off-by: dianew <dianew@vmware.com>
Start from VMware tools 12.0.0, there is new folder in VMware tools ISO package 'win10' for Windows 11 and Windows Server 2022. So here change to use pvscsi driver in VMware tools 11.3.5 'win8' folder for under testing guest OS to do installation on PVSCSI boot disk as a workaround.


keirazhang (approved):
LGTM. Thanks!

PR #331, opened 2022-06-23
Fix: lossed file extension

dentist128:
The systemd-resolve output is stored in /tmp/systemd_resolve_status.txt on the guest machine but at next step playbook tries to download /tmp/systemd_resolve_status (without extension) to controller host. This issue makes impossible to pass GOS tests on Ubuntu hosts.


Tomorrow9:
@dentist128 you can use "git commit -s -m 'description message'"  to submit with signoff info. Thanks.

PR #332, opened 2022-06-23
Fix: lost file extension

dentist128:
The systemd-resolve output is stored in /tmp/systemd_resolve_status.txt on the guest machine but at next step playbook tries to download /tmp/systemd_resolve_status (without extension) to controller host. This issue makes impossible to pass GOS tests on Ubuntu hosts.

### DIFF
--- a/linux/guest_customization/check_dns.yml
+++ b/linux/guest_customization/check_dns.yml
@@ -118,7 +118,7 @@
         - include_tasks: ../../common/vm_guest_file_operation.yml
           vars:
             operation: "fetch_file"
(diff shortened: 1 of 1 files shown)


dentist128:
The systemd-resolve output is stored in /tmp/systemd_resolve_status.txt on the guest machine but at next step playbook tries to download /tmp/systemd_resolve_status (without extension) to controller host. This issue makes impossible to pass GOS tests on Ubuntu hosts.

Tomorrow9 (approved):
Thanks for the contributing to this project and the fix.

PR #334, opened 2022-06-29
Fix Photon 3 OVA deploy failure and package installation failures

keirazhang:
This PR fixed below issues:
1. Failed to set root password with cloud-init at deploying Photon 3 OVA image. The issue has been fixed in Photon 3 shadow-4.6-6, but this fix worked around it for Photon 3 OVA images using shadow-4.6-5 or earlier. Meanwhile, added tasks for collecting cloud-init logs after deploying Photon and Ubuntu OVA for cloud-init problems triage.
2. Installing packages in Photon 3 might fail with error like "Couldn't resolve host name" or "Couldn't connect server" at the first time of executing `tdnf install`. Ultimately the root cause of the issue was this.
https://github.com/vmware/photon/commit/5b4f598f0ef30984217229c8bada56ff806e96eb
And fix is https://github.com/vmware/photon/commit/838af600bcb81a04de31af6b692e8e37c146d0ab

Photon team provides a workaround solution, which is to modify "/etc/systemd/resolved.conf" file to disable TLS options.
Like below:
```
...
DNSSEC=no
DNSOverTLS=no
...
```
And then run: "systemctl restart systemd-resolved"
3. Photon 4 uses sshd.socket as default sshd service for listening SSH connections, but Photon 3 or earlier uses sshd.service as default service. So for Photon 4 or later, it needs to start the sshd.socket after cloud-init at first boot to open the SSH port.

### DIFF
--- a/autoinstall/Photon/ks.cfg
+++ b/autoinstall/Photon/ks.cfg
@@ -7,7 +7,7 @@
     "disk": "/dev/{{ boot_disk_name }}",
     "packagelist_file": "packages_ova.json",
     "arch": "x86_64",
-    "additional_packages": ["vim","gawk","sudo"],
+    "additional_packages": ["vim","gawk","sudo","tar"],
     "install_linux_esx": true,
     "eject_cdrom": true,
     "postinstall": [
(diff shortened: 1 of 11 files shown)


Tomorrow9 (approved):
LGTM, thanks!

PR #335, opened 2022-07-04
[Windows] Comment out get VM IP address task after deploying MS Windows template

Tomorrow9:
Due to there are 2 network adapters available with network "VM Network", `vm_get_ip_from_vmtools.yml` returns error when try to get one available VM IP address, comment this task out for now since we'll not execute tasks in guest.


keirazhang (approved):
LGTM. Thanks~

PR #363, opened 2022-10-10
Fix Ubuntu sshd service name for setting TCPKeepAlive

keirazhang:
Ubuntu 22.10 changed sshd service name to ssh.service not sshd.service any more. This fix checked sshd.service existence firstly. If it doesn't exist, check ssh.service, and then use the existing one  to restart sshd service.
Besides, lsilogic_vhba_device_ops still fails with Ubuntu 22.10, so change the SCSI rescan method for both 22.04 and 22.10.

### DIFF
--- a/linux/utils/ssh_keep_alive.yml
+++ b/linux/utils/ssh_keep_alive.yml
@@ -28,27 +28,16 @@
       vars:
         service_name: "sshd"
 
-    - name: "Found {{ service_info.name }}, update sshd service info with it"
-      ansible.builtin.set_fact:
-        sshd_service_info: "{{ service_info }}"
-      when:
-        - service_info
-        - service_info.name is defined
-        - service_info.name
-        - service_info.state is defined
-        - service_info.state == "running"
-
     - name: "Not found sshd service, try to get ssh service info"
       include_tasks: get_service_info.yml
       vars:
         service_name: "ssh"
       when: >
-        service_info is undefined or
         not service_info or
         service_info.name is undefined or
         not service_info.name or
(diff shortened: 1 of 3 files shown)


linux/utils/ssh_keep_alive.yml
```diff
@@ -1,21 +1,78 @@
...
+        - service_info
+        - service_info.name is defined
+        - service_info.name
+        - service_info.state is defined
+        - service_info.state == "running"
+
+    - name: "Not found sshd service, try to get ssh service info"
+      include_tasks: get_service_info.yml
+      vars:
+        service_name: "ssh"
```
Tomorrow9, line 44: Here missed `when` condidtion?

linux/utils/ssh_keep_alive.yml
```diff
@@ -1,21 +1,78 @@
...
+      ansible.builtin.set_fact:
+        sshd_service_info: "{{ service_info }}"
+      when:
+        - service_info
+        - service_info.name is defined
+        - service_info.name
+        - service_info.state is defined
+        - service_info.state == "running"
+
+    - name: "Check sshd service exists"
```
Tomorrow9, line 56: "Check sshd or ssh service exists"

linux/utils/ssh_keep_alive.yml
keirazhang: yes, updated it.

Tomorrow9 (dismissed):
LGTM, thanks!

Tomorrow9 (approved):

PR #365, opened 2022-10-13
[Linux] Skip testing secureboot enable or disable for OS with vulnerable bootloader

keirazhang:
Skip secureboot testing for OS listed in https://kb.vmware.com/s/article/88737 when VM's hardware version >= 20.
We can't skip deploy_vm for these Linux guests as it requires ansible distribution information.


Tomorrow9 (approved):
LGTM, thanks!

PR #375, opened 2022-11-02
[Windows] Add check loaded inbox drivers after VMware Tools uninstall

Tomorrow9:
Signed-off-by: Diane Wang <dianew@vmware.com>


windows/wintools_uninstall_verify/verify_uninstall.yml
```diff
@@ -0,0 +1,34 @@
...
+# Get VMware Tools status after uninstall
+- include_tasks: ../../common/vm_get_vmtools_status.yml
+- name: "Verify VMware Tools is not running and not installed after uninstall"
+  ansible.builtin.assert:
+    that:
+      - vmtools_is_installed is defined
+      - not vmtools_is_installed | bool
+      - vmtools_is_running is defined
+      - not vmtools_is_running | bool
+    fail_msg: "Either VMware Tools status: '{{ vmtools_is_installed }}' is still installed or running: '{{ vmtools_is_running }}' after uninstall."
```
keirazhang, line 13: The fail message doesn't read smoothly.

windows/wintools_uninstall_verify/add_new_devices_check_status.yml
```diff
@@ -40,6 +22,19 @@
...
+- include_tasks: ../../common/vm_hot_add_remove_disk_ctrl.yml
+  vars:
+    disk_controller_ops: "present"
+    disk_controller_type: "paravirtual"
+- name: "Check add new pvscsi controller config changes"
+  ansible.builtin.assert:
+    that:
+      - disk_controller_facts is defined
+      - disk_controller_facts.changed
+    fail_msg: "Add new pvscsi controller task result 'changed' is not true."
```
keirazhang, line 35: Can this assert task be moved into `common/vm_hot_add_remove_disk_ctrl.yml`, so that every time device change can be checked.

Tomorrow9: Since there is no context in the common task, so we don't know if the task should be changed or not due to if there is existing controller, then the result is not changed.

keirazhang: Got it. Thanks~

keirazhang (approved):
LGTM. Thanks~

PR #380, opened 2022-11-11
[Windows][Bug] Fix disk UniqueId changed issue after OS restart

Tomorrow9:
Signed-off-by: Diane Wang [dianew@vmware.com](mailto:dianew@vmware.com)
1. disk UniqueId will be changed after OS restart on ESXi 7.0.3.
2. disk number will be changed after OS restart on ESXi 7.0.0/7.0.1/8.0.0.

So here changed to if can not get file content in new disk, check all disk status and change offline status to online.


keirazhang (approved):
LGTM. Thanks!

PR #390, opened 2022-12-02
change the known issue condition of VMware Tools version

123lzxm:
change the known issue condition of VMware Tools version for case uninstall tools. 

Signed-off-by: Yanan Shen <yanans@vmware.com>


Tomorrow9 (approved):
LGTM, thanks

PR #414, opened 2023-01-17
Fix regression issue found in SLES/SLED/Ubuntu

keirazhang:
There is no default GPG key for SLES/SLED, so remove GPG check for adding SLES/SLED repos. 
Ubuntu 20.04.5 server failed to ndctl, rdma packages during deploy_vm, so in this changes only Ubuntu Server >= 22.04 will install these packages at deploy_vm. Ubuntu 20.04.5 will install these packages in test cases.


Tomorrow9 (approved):
LGTM, thanks!

123lzxm (approved):
LGTM, thanks.

PR #419, opened 2023-02-06
[Linux] Improve check_os_fullname test case to check guest id and guest family

keirazhang:
The improvement in this change 
1. identified different behavior when AlmaLinux/Rocky Linux 8.x&9.x has different VMware Tools version on different ESXi version.
2. added guest id and guest family checking


linux/check_os_fullname/validate_os_fullname.yml
```diff
@@ -1,63 +1,95 @@
...
+              Guest id is '{{ guestinfo_guest_id }}' on ESXi {{ esxi_version }}, and
+              guest OS full name is '{{ guestinfo_guest_full_name }}',  which shows
+              guest OS detailed information and is as expected on ESXi {{ esxi_version }}.
+              Test Passed.
+        - name: "Test passed on ESXi {{ esxi_version }}"
+          ansible.builtin.meta: end_host
+      when:
+        - unmapped_guest_fullname is defined and unmapped_guest_fullname
+        - unmapped_guest_fullname in guestinfo_guest_full_name
+  when: >
```
123lzxm, line 39: It seems here is missing one parameter when guest OS is otherlinux

linux/deploy_vm/deploy_vm.yml
```diff
@@ -36,6 +36,10 @@
     # Get SSH public key from localhost
     - include_tasks: ../utils/get_local_ssh_public_key.yml
 
+    - name: "Set fact of vm_exists to False"
```
123lzxm, line 39: vm_exisits will be set in env_setup. Why add such code here?

linux/check_os_fullname/validate_os_fullname.yml
keirazhang: There is no OS in our [compatible GOS list](https://github.com/vmware/ansible-vsphere-gos-validation#compatible-guest-operating-systems) will be mapped to `otherLinuxGuest` or `otherLinux64Guest`. If it does, it should be a product bug.

linux/deploy_vm/deploy_vm.yml
keirazhang: You are right. Will remove it.

linux/check_os_fullname/validate_os_fullname.yml
123lzxm: I'm still confused about the logic here. E.g. if we run the case with rhe9 on ESXi 7.0.0, will this block be executed?

linux/check_os_fullname/validate_os_fullname.yml
```diff
@@ -1,63 +1,95 @@
...
+# or on ESXi 7.0.0, that's expected. Test passed.
+- name: "Check unmapped guest OS full name"
+  block:
+    # The unmapped guest full name which at least includes Linux kernel version, OS pretty name
+    # For VM with newer VMware Tools, the unmapped guest full name also includes OS name, version id and build id
+    # Here we only check linux kernel version and OS pretty name
+    - name: "Set unmapped guest full name for {{ vm_guest_os_distribution }} on ESXi {{ esxi_version }}"
+      ansible.builtin.set_fact:
+        unmapped_guest_fullname: "Linux {{ guest_os_ansible_kernel }} {{ guest_os_release.PRETTY_NAME }}"
+      when: guest_os_release.PRETTY_NAME | default('')
```
Tomorrow9, line 23: Missed condition "guest_is_otherlinux" is true and "expected_guest_fullname" is still ""?

keirazhang: Moved checking expected_guest_id and expected_guest_fullname to validate_os_fullname.yml

Tomorrow9: I mean does this whole block of checking unmapped guest fullname miss when condition?

Tomorrow9 (approved):
LGTM, thanks!

PR #420, opened 2023-02-06
change msg.autoanswer to msg.autoAnswer

123lzxm:
Signed-off-by: Yanan Shen <yanans@vmware.com>

### DIFF
--- a/common/vm_configure_cdrom.yml
+++ b/common/vm_configure_cdrom.yml
@@ -61,7 +61,7 @@
             vm_advanced_settings:
               - key: "cdrom.showIsoLockWarning"
                 value: "FALSE"
-              - key: "msg.autoanswer"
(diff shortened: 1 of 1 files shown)


keirazhang:
Both "msg.autoanswer" and "msg.autoAnswer" should be fine. You can compare the lower case of extra config name in common/vm_set_extra_config.yml.

123lzxm:
> Both "msg.autoanswer" and "msg.autoAnswer" should be fine. You can compare the lower case of extra config name in common/vm_set_extra_config.yml.

Thanks for your review.  I thinks it's better to keep one format. Since "msg.autoanswer" does not work on main branch, so we prefer to use "msg.autoAnswer" .

Tomorrow9 (approved):
LGTM, thanks!

PR #443, opened 2023-03-24
[FreeBSD ]modify README file to add note about freeBSD auto-install

ZouYuhua:
modify README file to add note about freeBSD auto-install


keirazhang (approved):
LGTM. Thanks!

PR #452, opened 2023-04-06
Support VMware Photon OS 5.0 automation tests

keirazhang:
1. Add user-data and meta-data collection in deploy_vm.
2. Fixed photon.repo error for installing packages

### DIFF
--- a/linux/utils/add_official_online_repo.yml
+++ b/linux/utils/add_official_online_repo.yml
@@ -183,12 +183,12 @@
   block:
     - name: "Set the fact of VMware Photon OS {{ guest_os_ansible_distribution_major_ver }} online repositories"
       ansible.builtin.set_fact:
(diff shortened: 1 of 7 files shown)


linux/utils/add_official_online_repo.yml
```diff
@@ -200,6 +200,15 @@
         option_value: 1
       with_items: "{{ photon_online_repos }}"
 
+    - name: "Disable VMware Photon OS online repository - photon.repo"
```
123lzxm, line 203: Is the repo photon.repo enabled by defalut for Photon OS 5?
If not, no need to disable it here, since from line 191, only two repos  photon-release and photon-updates will be enabled when photon OS >=4.

keirazhang: Yes, it is enabled by default.

123lzxm (approved):
LGTM, thanks.

PR #454, opened 2023-04-07
Set VM primary NIC MAC address at env_setup or deploy_vm

keirazhang:
We have a limitation about testing VM must have only 1 network adapter and it should be connected. That means we can get VM's primary NIC at env_setup for existing VM or deploy_vm for new VM. So it can simplify the process of getting VM guest IP of the primary NIC. Files `vm_get_primary_nic.yml` and `vm_test_nic.yml` can be removed.


common/vm_get_ip.yml
```diff
@@ -2,30 +2,42 @@
 # SPDX-License-Identifier: BSD-2-Clause
 ---
 # Get VM IP Address using esxcli command or from VMware Tools guestinfo
-- name: Initialize VM guest IP address variable
+# Parameter:
+#   vm_primary_nic_mac: The VM's primary network adapter MAC address must be set
+# Return:
+#   vm_guest_ip: The VM's guest IPv4 address
+#
+- name: "Check vm_primary_nic_mac is set"
```
Tomorrow9, line 10: This check also need to be added in tasks `vm_get_ip_from_notools.yml`, `vm_get_ip_from_vmtools.yml` in case they are called separately.

keirazhang: Parameter checking was added in the two sub task files. But it is not recommended to call them directly as vm_get_ip.yml would check VM's power state and power on VM if necessary, the two sub tasks didn't check power state.

Tomorrow9 (approved):
LGTM, thanks!

PR #459, opened 2023-04-14
[Windows] Fix variables undefined error during deploying VM

123lzxm:
1. Getting the guest OS build number before enabling VBS
2. Do not add key provider when key provider type is not defined.

### DIFF
--- a/windows/deploy_vm/deploy_vm_from_iso.yml
+++ b/windows/deploy_vm/deploy_vm_from_iso.yml
@@ -126,7 +126,6 @@
   block:
     - name: "Get guest OS system info"
       include_tasks: ../utils/get_windows_system_info.yml
-      when: guest_os_system_info_retrieved is undefined or not guest_os_system_info_retrieved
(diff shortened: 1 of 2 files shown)


windows/deploy_vm/deploy_vm_from_iso.yml
```diff
@@ -124,6 +124,9 @@
 
 - name: "Enable VBS"
   block:
+    - name: "Get guest OS system info"
+      include_tasks: ../utils/get_windows_system_info.yml
+      when: guest_os_system_info_retrieved is undefined or not guest_os_system_info_retrieved
```
keirazhang, line 129: Looks like it is not necessary to add when here because this is a new deployed VM.

123lzxm: Yes. Remove the condition. Thanks .

keirazhang (approved):
LGTM. Thanks!

Issue #397, opened 2022-12-23
[Bug] move check_vm_settings.yml to be executed after revert to base snasphot


Tomorrow9:
##### SUMMARY

`check_vm_settings.yml` is executed in `env_setup`, the VM state in this stage is not the one run testing, testing will fail if VM is not in the base snapshot (clean) state. So we can consider moving this task file after reverting VM base snapshot.

##### ISSUE TYPE
- Bug Report

##### ANSIBLE VERSION

```paste below
"ansible [core 2.13.4]",
        "  config file = /home/worker/workspace/Ansible_Windows_Server_LTSC_Physical_main_LSILOGICSAS_E1000E_EFI/ansible-vsphere-gos-validation/ansible.cfg",
        "  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']",
        "  ansible python module location = /usr/lib/python3.9/site-packages/ansible",
        "  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections",
        "  executable location = /usr/bin/ansible",
        "  python version = 3.9.1 (default, Aug 19 2021, 02:58:42) [GCC 10.2.0]",
        "  jinja version = 3.0.1",
        "  libyaml = True"
```

##### CONFIGURATION

```paste below
```

##### ENVIRONMENT / OS

All

##### EXPECTED RESULTS

Testing not fail since if VM revert to the base snapshot then the checking will pass.

##### ACTUAL RESULTS

Testing failed.

keirazhang:
Hi Yuhua,
We need to make some changes in env_setup.yml for existing VM:
1. Set default value in vars/test.yml for revert_to_snapshot to 'BaseSnapshot', and uncomment this var and 'base_snapshot_name'
2. Move condition `when: cleanup_old_snapshots | bool` into `common/vm_cleanup_snapshot.yml` for `include_tasks: vm_remove_snapshot.yml` only.
3. Move down `- include_tasks: ../common/vm_get_vm_info.yml` after `check_vm_settings.yml`.

PR #460, opened 2023-04-18
Check start_connected is true when MV is powered off at env_setup

keirazhang:
When VM is powered off at env_setup, its network adapter's connected is false. So we need to check its start_connected is true so that it can be connected after power on.

### DIFF
--- a/env_setup/check_vm_settings.yml
+++ b/env_setup/check_vm_settings.yml
@@ -4,6 +4,9 @@
 # Check VM's settings which only has one network adapter and it
 # should be connected to VM.
 #
+- name: "Get VM's power state"
+  include_tasks: ../common/vm_get_power_state.yml
+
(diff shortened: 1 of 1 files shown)


env_setup/check_vm_settings.yml
```diff
@@ -4,14 +4,18 @@
 # Check VM's settings which only has one network adapter and it
 # should be connected to VM.
 #
+- name: "Get VM's power state"
+  include_tasks: ../common/vm_get_power_state.yml
+
 - name: "Get VM's network adapters info"
```
123lzxm, line 10: Has got the network adapter info for the existing VM in env_setup.yml, see line 38. We can save the adapter info there and ues it here.

keirazhang:
This issue can be fixed together with #397. So close this PR.

PR #469, opened 2023-06-05
Check VMware Tools version after OVA/OVF deployment without poweron

ZouYuhua:
From John:
Can you please check to see if our testing of OVAs includes verification of vmdk that ddb.toolsVersion is set to 2147483647 if open-vm-tools is included?  If we are not doing this then please add as a candidate new test.  I think the check needs to be done without booting the VM to be valid.

Test result: 
1) test with ubuntu 23.04 OVA:
image

2) test with photon 5.0 OVA
image

### DIFF
--- a/common/collect_ovf_vm_config.yml
+++ b/common/collect_ovf_vm_config.yml
@@ -249,3 +249,13 @@
   ansible.builtin.copy:
     dest: "{{ ovf_vm_hardware_config_path }}/{{ ovf_vm_hardware_config_file }}"
     content: "{{ [ovf_vm_hardware_config] | to_nice_json }}"
+
+- name: "Display ddb.toolsVersion of open-vm-tools in OVA"
+  debug: var=ovf_vm_vmtools_info['Config Tools Version']
+        
+- name: "Check ddb.toolsVersion of open-vm-tools in OVA is expected"
+  ansible.builtin.assert:
+    that:
+      - ovf_vm_vmtools_info['Config Tools Version'] | length > 0 or vm_extra_config['guestinfo.vmtools.versionNumber'] | length > 0
+      - ovf_vm_vmtools_info['Config Tools Version'] == '2147483647' or ovf_vm_vmtools_info['Config Tools Version'] == vm_extra_config['guestinfo.vmtools.versionNumber']
+    fail_msg: "The ddb.toolsVersion in OVA template is {{ ovf_vm_vmtools_info['Config Tools Version'] }}, not as expected '2147483647' or {{ vm_extra_config['guestinfo.vmtools.versionNumber'] }}."
--- a/common/collect_ovf_vm_config.yml
+++ b/common/collect_ovf_vm_config.yml
@@ -250,12 +250,10 @@
     dest: "{{ ovf_vm_hardware_config_path }}/{{ ovf_vm_hardware_config_file }}"
     content: "{{ [ovf_vm_hardware_config] | to_nice_json }}"
 
-- name: "Display ddb.toolsVersion of open-vm-tools in OVA"
-  debug: var=ovf_vm_vmtools_info['Config Tools Version']
-        
 - name: "Check ddb.toolsVersion of open-vm-tools in OVA is expected"
   ansible.builtin.assert:
     that:
       - ovf_vm_vmtools_info['Config Tools Version'] | string | length > 0 or vm_extra_config['guestinfo.vmtools.versionNumber'] | length > 0
       - ovf_vm_vmtools_info['Config Tools Version'] | string in ['2147483647', vm_extra_config['guestinfo.vmtools.versionNumber']]
     fail_msg: "The ddb.toolsVersion in OVA template is {{ ovf_vm_vmtools_info['Config Tools Version'] }}, not as expected '2147483647' or {{ vm_extra_config['guestinfo.vmtools.versionNumber'] }}."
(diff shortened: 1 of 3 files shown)


keirazhang:
We already have a common task `common/collect_ovf_vm_config.yml`, in which ddb.toolsVersion is collected from `vm_config.config.tools.toolsVersion`.
```
- name: "Get VMware Tools info"
  ansible.builtin.set_fact:
    ovf_vm_vmtools_info: >-
      {{  
        ovf_vm_vmtools_info |
        combine({'Version': vm_config.config.tools.toolsVersion | default(''),
                 'Install Type': vm_config.config.tools.toolsInstallType | default(''),
                 'Upgrade Policy': vm_config.config.tools.toolsUpgradePolicy | default(''),
                 'Sync Time with Host': vm_config.config.tools.syncTimeWithHost | default('')})
      }}  

 I think you can directly use it with some changes:
1. Move out included task `vm_get_vm_info.yml` from `common/collect_ovf_vm_config.yml`, and replace vars like `vm_guest_id` in `common/collect_ovf_vm_config.yml` with `vm_config.config.guestId`.
2. The moved out `vm_get_vm_info.yml` can be put after  `common/collect_ovf_vm_config.yml` in `linux/deploy_vm/deploy_vm_from_ova.yml` and `windows/deploy_vm/deploy_vm_from_ova.yml`
3. Move the power on task `common/vm_set_power_state.yml` from `deploy_vm_from_ova.yml` into `common/collect_ovf_vm_config.yml`, and put it before `vm_wait_vmtools_status.yml`
4. In `common/collect_ovf_vm_config.yml`, it collected vmtools version twice: one from vm_config.config.tools.toolsVersion and the other one from guestinfo_vmtools_info. And the second one will overwrite the first one.  Since we need to check ddb.toolsVersion before power-on, I suggest you save them with two different keys in `ovf_vm_vmtools_info`. For example, we can save the first one as 'Config Tools Version' and the second one as 'Guest Tools Version'. 
5. The default ddb.toolsVersion is 2147483647, but OS vendor might change it to real tools version number in meta data. So you need to check it is either 2147483647 or same as `guestinfo.vmtools.versionNumber` get from `vm_extra_config`.

- 

```

linux/deploy_vm/deploy_vm_from_ova.yml
```diff
@@ -56,15 +56,25 @@
...
+      block:
+        - name: "Display toolsVersion of open-vm-tools in OVA"
+          debug: var=ovf_vm_vmtools_info['Config Tools Version']
+        
+        - name: "Check toolsVersion of open-vm-tools in OVA is expected"
+          ansible.builtin.assert:
+            that:
+              - ovf_vm_vmtools_info['Config Tools Version'] in ['2147483647', vm_extra_config['guestinfo.vmtools.versionNumber']]
+            fail_msg: "toolsVersion of open-vm-tools in OVA is not expected"
+      when: ovf_vm_vmtools_info['Guest Tools Version'] is match("build-")
```
keirazhang, line 77: You can put this checking to the end of `common/collect_ovf_vm_config.yml`, so that both linux and windows OVA can be checked.

In the assertion, you need to firstly check ovf_vm_vmtools_info['Config Tools Version'] or vm_extra_config['guestinfo.vmtools.versionNumber'] is not empty. And
please change the fail message be more specific. For example,
```
fail_msg: "The ddb.toolsVersion in OVA template is {{ ovf_vm_vmtools_info['Config Tools Version'] }}, not as expected '2147483647' or {{ vm_extra_config['guestinfo.vmtools.versionNumber']] }}.
```

The when condition is not needed

common/collect_ovf_vm_config.yml
```diff
@@ -247,3 +249,13 @@
   ansible.builtin.copy:
     dest: "{{ ovf_vm_hardware_config_path }}/{{ ovf_vm_hardware_config_file }}"
     content: "{{ [ovf_vm_hardware_config] | to_nice_json }}"
+
+- name: "Display ddb.toolsVersion of open-vm-tools in OVA"
+  debug: var=ovf_vm_vmtools_info['Config Tools Version']
```
keirazhang, line 254: You can remove this debug message but add a "success_msg" to below assert task, like:
```
success_msg: "The The ddb.toolsVersion in OVA template is {{ ovf_vm_vmtools_info['Config Tools Version'] }}, which is as expected."
```

windows/deploy_vm/deploy_vm_from_ova.yml
```diff
@@ -41,12 +41,6 @@
...
     vm_exists: true
 
-# Power on VM and wait for guest fullname reported by VMware tools
-- name: "Power on VM"
-  include_tasks: ../../common/vm_set_power_state.yml
-  vars:
-    vm_power_state_set: "powered-on"
-
 - name: "Get VM info"
   include_tasks: ../../common/vm_get_vm_info.yml
```
keirazhang, line 45: Move `vm_get_vm_info.yml` after `collect_ovf_vm_config.yml` as it must be executed after VM power-on for retrieving primary nic MAC.

ZouYuhua: Change it. Thanks

common/collect_ovf_vm_config.yml
ZouYuhua: Change it. Thanks

linux/deploy_vm/deploy_vm_from_ova.yml
ZouYuhua: Change it. Thanks

keirazhang (approved):
LGTM. Thanks!

PR #476, opened 2023-06-19
Add parameter esxi_hostname to vm_create

ZouYuhua:
With multiple clusters under same datacenter, we need to add parameter "esxi_hostname" or "cluster name" to common task vm_create.yml to specify the the esxi_hostname/cluster. Here uses `esxi_hostname` for creating VM.


keirazhang (approved):
LGTM. Thanks!

PR #479, opened 2023-07-07
[Linux] Fix cloud-init GOSC failure on Ubuntu and Photon OS

keirazhang:
When enabling and starting cloud-init services on Ubuntu with latest cloud-init version, it will cause network interface down. By consulting with GOSC team, it is suggested to only enable but no start cloud-init services when enabling cloud-init GOSC.

On Photon OS OVA deployment, installing latest chrony will also lead to cloud-init service failure. As chrony won't affect Photon OS testing, so this fix removed it from Photon OS OVA's user data.

Besides, to gain more information to debug GOSC failures, this fix adjusted GOSC testing workflow, which will always collect tools deploypkg log for both perl GOSC and cloud-init, cloud-init config and logs for cloud-init GOSC.

And some ansible-lint errors were also fixed.

### DIFF
--- a/linux/deploy_vm/templates/photon-ova-user-data.j2
+++ b/linux/deploy_vm/templates/photon-ova-user-data.j2
@@ -5,7 +5,6 @@ ssh_authorized_keys:
 packages:
   - sudo
   - sg3_utils
-  - chrony
   - tar
   - ndctl
   - python3-rpm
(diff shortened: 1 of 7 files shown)


ZouYuhua (commented):
LGTM.  Thanks

ZouYuhua (approved):

PR #486, opened 2023-08-04
Ignore esxi_hostname at VM creation when vcenter_hostname is not defined

keirazhang:
If there is no vcenter_hostname defined in vars/test.yml, vsphere_host_name will be same with esxi_hostname, and `common/vm_create.yml` will report error like "Unable to find host". It means that esxi_hostname is required when vsphere_host_name = vcenter_hostname for this task. The solution is to ignore esxi_hostname when vsphere_host_name = esxi_hostname.


ZouYuhua:
LGTM. @keirazhang

123lzxm (approved):
LGTM. Thanks.

PR #477, opened 2023-06-26
Support RHEL CoreOS automation testing

keirazhang:
RHCOS is very like Flatcar. It uses Ignition to customize guest OS after OVA deployment, and for test cases not supported by Flatcar, RHCOS doesn't support either.  Besides, Flatcar's ansible distribution name has been fixed in ansible-core 2.14 (https://github.com/ansible/ansible/pull/77635), so now we can directly use guest_os_ansible_distribution is 'Flatcar' or not to check it. Some when conditions including 'Flatcar' checking are simplified in this fix.

```
VM information:
+-----------------------------------------------------------------+
| Name                      | test_rhcos                          |
+-----------------------------------------------------------------+
| Guest OS Distribution     | RHCOS 4.13 x86_64                   |
+-----------------------------------------------------------------+
| IP                        | 10.78.182.158                       |
+-----------------------------------------------------------------+
| GUI Installed             | False                               |
+-----------------------------------------------------------------+
| CloudInit Version         |                                     |
+-----------------------------------------------------------------+
| Hardware Version          | vmx-21                              |
+-----------------------------------------------------------------+
| VMTools Version           | 12.1.5 (build-20735119)             |
+-----------------------------------------------------------------+
| Config Guest Id           | rhel8_64Guest                       |
+-----------------------------------------------------------------+
| GuestInfo Guest Id        | rhel8_64Guest                       |
+-----------------------------------------------------------------+
| GuestInfo Guest Full Name | Red Hat Enterprise Linux 8 (64-bit) |
+-----------------------------------------------------------------+
| GuestInfo Guest Family    | linuxGuest                          |
+-----------------------------------------------------------------+
| GuestInfo Detailed Data   |                                     |
+-----------------------------------------------------------------+


Test Results (Total: 30, Passed: 21, Failed: 1, Skipped: 8, Elapsed Time: 01:11:21)
+-------------------------------------------------------------------------+
| ID | Name                                 |   Status        | Exec Time |
+-------------------------------------------------------------------------+
| 01 | deploy_vm_ova                        |   Passed        | 00:04:56  |
| 02 | check_inbox_driver                   |   Passed        | 00:02:18  |
| 03 | ovt_verify_install                   | * Not Supported | 00:01:02  |
| 04 | ovt_verify_status                    |   Passed        | 00:01:10  |
| 05 | vgauth_check_service                 |   Passed        | 00:00:53  |
| 06 | check_ip_address                     |   Passed        | 00:00:42  |
[...]

PR #489, opened 2023-08-10
[Linux] Shutdown guest OS after auto install completes instead of rebooting

keirazhang:
Even though guest OS sends autoinstall complete message to serial port, it might still be running some post tasks and it won't be rebooted immediately. Sometimes the post tasks take some time, which led to ansible tasks thought starting to wait for guest full name but it failed due to guest OS hasn't reboot at all. 

It is difficult to find the best timing when to wait for guest full name, so in this change I replaced 'reboot' with 'shutdown' as the following action after autoinstall is compeleted. Therefore, we only needs to wait for VM's power states becoming poweredOff after autoinstall complete message received. Then we can consider the complete autoinstall process is finished, and then power on VM to wait for its guest full name.

In this change, I also added boot options for redirecting console output to serial port, so we can have more information when debugging deploy_vm failures.

### DIFF
--- a/linux/deploy_vm/delete_unattend_install_iso.yml
+++ b/linux/deploy_vm/delete_unattend_install_iso.yml
@@ -1,17 +1,8 @@
-# Copyright 2021-2023 VMware, Inc.
+# Copyright 2023 VMware, Inc.
 # SPDX-License-Identifier: BSD-2-Clause
 ---
-# Eject CDROM of unattend install ISO in guest OS
-#- include_tasks: ../utils/eject_cdrom_in_guest.yml
-#  vars:
-#    guest_cdrom_device_num: "{{ os_install_iso_list_len | int -1 }}"
-#  when: guest_os_family != "FreeBSD"
+# Delete unattend install ISO file
 #
-#- name: "Eject CD for FreeBSD"
-#  ansible.builtin.command: "camcontrol eject /dev/cd0"
-#  delegate_to: "{{ vm_guest_ip }}"
-#  when: guest_os_family == "FreeBSD"
-
 # By default, it's attached to the last CDROM in composed VM CDROMs list
 - name: "Set fact of the controller and unit number of CDROM attaching unattend install ISO"
   ansible.builtin.set_fact:
--- a/linux/deploy_vm/deploy_vm_from_iso.yml
+++ b/linux/deploy_vm/deploy_vm_from_iso.yml
@@ -195,6 +195,14 @@
         - unattend_install_conf is defined
         - ('UOS' in unattend_install_conf)
 
+    # When there is "reboot: Power down" in serial port log, it means the auto install is completed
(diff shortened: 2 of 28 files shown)


linux/deploy_vm/deploy_vm_from_iso.yml
```diff
@@ -213,12 +203,80 @@
...
+        vm_take_screenshot_local_path: "{{ current_test_log_folder }}"
+
+    # VMware Photon OS kickstart file can't add shutdown action,
+    # so here needs to shutdown it separately
+    - name: "Shutdown VMware Photon OS VM after auto install completes"
+      when:
+        - unattend_install_conf is defined
+        - unattend_install_conf is match('Photon')
+      block:
+        - name: "Wait for guest full name on VMware Photon OS VM"
```
123lzxm, line 218: Why get full name here? Does it need to be executed after power off/power on?

linux/deploy_vm/delete_unattend_install_iso.yml
```diff
@@ -1,17 +1,8 @@
-# Copyright 2021-2023 VMware, Inc.
+# Copyright 2023 VMware, Inc.
 # SPDX-License-Identifier: BSD-2-Clause
 ---
-# Eject CDROM of unattend install ISO in guest OS
```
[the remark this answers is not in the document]
keirazhang: Yes, because now OS will be shutdown after autoinstall completes. And this task is executed after shutdown which means VM is powered off at that time.

linux/deploy_vm/deploy_vm_from_iso.yml
keirazhang: Photon OS won't shutdown or reboot after autoinstall complete. Here adds a task to wait for guest full name is to ensure the guest OS is completely up as tools can collect guest full name successfully.

linhuali (approved):
LGTM. Thanks.

linux/deploy_vm/deploy_vm_from_iso.yml
```diff
@@ -213,12 +203,80 @@
...
+        - transferred_unattend_iso is defined
+        - transferred_unattend_iso
+
+    - name: "Change the boot disk to be first boot device in boot order"
+      include_tasks: ../../common/vm_set_boot_options.yml
+      vars:
+        boot_order_list: ['disk']
+      when: guest_id is match('ubuntu.*')
+
+    - name: "Download serial output file before removing serial port"
```
123lzxm, line 247: If request #477 is merged, here can use collect_serial_port_log.yml adeed by #477

keirazhang: If there is no more comments about this change. I will firstly merge this one, and then update here in https://github.com/vmware/ansible-vsphere-gos-validation/pull/477 together with other unresolved comments.

PR #496, opened 2023-08-31
[Pardus] the offical repo is changed for pardus 23.0

ZouYuhua:
Issue:
The official repo is changed in pardus 23.0 which cause the failure to install packages during autoinstall.

Test result:
image


keirazhang (approved):
LGTM. Thanks!

PR #498, opened 2023-09-07
[Linux] Check vmtools_log_dir is defined before using in check_os_fullname

keirazhang:
When test failed at test_setup, it won't define `vmtools_log_dir`, but at always block, this var is used. So here needs to check `vmtools_log_dir` is defined.

### DIFF
--- a/linux/check_os_fullname/check_os_fullname.yml
+++ b/linux/check_os_fullname/check_os_fullname.yml
@@ -102,29 +102,34 @@
         - name: "Test rescue"
           include_tasks: ../../common/test_rescue.yml
       always:
-        - name: "Get VMware Tools log directory stat info"
-          include_tasks: ../utils/get_file_stat_info.yml
-          vars:
-            guest_file_path: "{{ vmtools_log_dir }}"
-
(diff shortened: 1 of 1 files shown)


linux/check_os_fullname/check_os_fullname.yml
```diff
@@ -102,29 +102,34 @@
...
+                  include_tasks: ../utils/fetch_file.yml
+                  vars:
+                    fetch_file_src_path: "{{ vmtools_log_dir }}.tgz"
+                    fetch_file_dst_path: "{{ current_test_log_folder }}/"
+                    fetch_file_ignore_errors: true
+                  when:
+                    - archive_vmtools_logs is defined
+                    - archive_vmtools_logs.changed is defined
+                    - archive_vmtools_logs.changed
+              when: guest_file_exists
```
123lzxm, line 135: The block condition can move to line 116

keirazhang: Nice catch. Thanks!

123lzxm (approved):
LGTM, thanks.

PR #500, opened 2023-09-12
[FreeBSD 14] The default nvme disk name is changed in FreeBSD 14 which cause the failure for autoinstall

ZouYuhua:
The nvme disk name in FreeBSD 13 is nvd0, but in FreeBSD it is changed to nda0.
We need to change autoinstall for FreeBSD 14.

Test result:
image

### DIFF
--- a/autoinstall/FreeBSD/installerconfig
+++ b/autoinstall/FreeBSD/installerconfig
@@ -9,7 +9,7 @@ export ZFSBOOT_CONFIRM_LAYOUT=0
 
 {% set freebsd_major_version = 14 %}
 {% if guest_os_type is defined and guest_os_type %}
-{% set freebsd_major_version = `echo $guest_os_type | cut -d- -f2 | cut -d. -f1` %}
+{% set freebsd_major_version = guest_os_type.split('-')[1].split('.')[0] %}
 {% endif %}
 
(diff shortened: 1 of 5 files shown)


ZouYuhua:
@keirazhang  Please help codereview it.  Thanks

autoinstall/FreeBSD/installerconfig
```diff
@@ -7,7 +7,14 @@ export ZFSBOOT_SWAP_SIZE=2g
 export nonInteractive="YES"
 export ZFSBOOT_CONFIRM_LAYOUT=0
 
-{% if boot_disk_controller == 'nvme' %}
+{% set freebsd_major_version = 14 %}
+{% if guest_os_type is defined and guest_os_type %}
```
keirazhang, line 11: where did you define `guest_os_type`? If there is no good way to identify FreeBSD version before deploying vm. I think you can add a new installerconfig for FreeBSD 14. Move the original one as `autoinstall/FreeBSD/13/installerconfig`, and add a new one for FreeBSD 14: `autoinstall/FreeBSD/14/installerconfig`. The only difference between the two installerconfig is the value of `ZFSBOOT_DISKS`.

ZouYuhua:
@keirazhang 
Passed after code change:
image

Thanks

keirazhang (approved):
LGTM. Thanks!

PR #504, opened 2023-09-25
[Linux] Fix undefined variables at rescue or always block when test_setup failed

keirazhang:
If test case fails at test_setup, some variables used in rescue or always block may be not defined but used, which would lead to variable undefined error. This fix is to add conditions to use those variables.

### DIFF
--- a/linux/check_inbox_driver/check_inbox_driver.yml
+++ b/linux/check_inbox_driver/check_inbox_driver.yml
@@ -65,7 +65,8 @@
           when: guest_os_ansible_distribution == "FreeBSD"
 
       rescue:
-        - include_tasks: ../../common/test_rescue.yml
+        - name: "Test case failure"
+          include_tasks: ../../common/test_rescue.yml
       always:
         - name: "Dump inbox drivers versions into a json file"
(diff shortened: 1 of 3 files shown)


linhuali (dismissed):
LGTM. Thanks!

linux/check_inbox_driver/check_inbox_driver.yml
```diff
@@ -59,31 +59,36 @@
...
           include_tasks: get_inbox_drivers.yml
           when: guest_os_ansible_distribution != "FreeBSD"
-        
+
         - name: "Get FreeBSD inbox drivers"
           include_tasks: get_freebsd_inbox_drivers.yml
           when: guest_os_ansible_distribution == "FreeBSD"
 
       rescue:
         - include_tasks: ../../common/test_rescue.yml
```
123lzxm, line 68: Add name?

keirazhang: Done. Thanks!

123lzxm (approved):
LGTM, thanks.

PR #512, opened 2023-11-13
Increase time out for waiting Linux GOSC finished message

keirazhang:
Cloud-init GOSC with static IP address on Ubuntu 23.10 took more than 100s to finish. So this change increased the default timeout to 300s.


Tomorrow9 (approved):
LGTM, thanks

PR #526, opened 2023-12-25
[Windows] Fix test case 'gosc_sanity_dhcp' failure

Tomorrow9:
1. always update host inventory after GOSC even DHCP IP is not changed.
2. save back Administrator user password after GOSC test cases.


keirazhang (approved):
LGTM. Thanks!

123lzxm (approved):
LGTM, thanks.

PR #535, opened 2024-01-18
Add current_test_timestamp at test_setup as a common variable

keirazhang:
Added a new variable `current_test_timestamp` for each test case, which can be used to replace expressions like `{{ lookup('pipe', 'date +%Y-%m-%d-%H-%M-%S') }}`.


Tomorrow9 (dismissed):
LGTM, thanks!

Tomorrow9 (approved):

PR #550, opened 2024-02-21
Change conditions of IP address checking in vm_get_ip.yml

Tomorrow9:


common/vm_get_ip_esxcli.yml
```diff
@@ -28,6 +28,7 @@
   until:
     - get_vm_ip_esxcli.stdout is defined
     - get_vm_ip_esxcli.stdout
+    - get_vm_ip_esxcli.stdout | from_yaml
```
123lzxm, line 31: The reason for adding the condition?

123lzxm (dismissed):

common/vm_get_ip.yml
```diff
@@ -48,6 +48,8 @@
 - name: "Check VM '{{ vm_name }}' IP address"
   ansible.builtin.assert:
     that:
-      - vm_guest_ip is defined and vm_guest_ip
-    fail_msg: "Failed to get VM '{{ vm_name }}' IP Address"
+      - vm_guest_ip
+      - vm_guest_ip | ansible.utils.ipaddr
+      - vm_guest_ip != "0.0.0.0"
```
keirazhang, line 53: 169.254.x.x is not a valid IP either.

keirazhang (approved):

common/vm_get_ip.yml
Tomorrow9: Add it, thanks.

common/vm_get_ip_esxcli.yml
Tomorrow9: That line can be removed. Thanks.

PR #560, opened 2024-03-25
[Windows] Remove old base snapshot when reset base snapshot

Tomorrow9:
Old base snapshot is not used anymore when testing proceeds and new base snapshot is created.

### DIFF
--- a/windows/wintools_complete_install_verify/wintools_complete_install_verify.yml
+++ b/windows/wintools_complete_install_verify/wintools_complete_install_verify.yml
@@ -128,6 +128,8 @@
 
         - name: "Reset base snapshot"
           include_tasks: ../../common/reset_base_snapshot.yml
+          vars:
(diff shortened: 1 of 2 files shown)


windows/wintools_complete_install_verify/wintools_complete_install_verify.yml
```diff
@@ -128,6 +128,8 @@
 
         - name: "Reset base snapshot"
           include_tasks: ../../common/reset_base_snapshot.yml
+          vars:
```
123lzxm, line 131: Linux has the same scenario. It's also can do the same change. Or create another merge request for Linux?

keirazhang: No change for Linux. Thanks!

keirazhang (approved):
LGTM. Thanks!

PR #554, opened 2024-03-04
Display ESXi model, CPU model and code name at deploy_vm failure

keirazhang:
Added a script to query Intel CPU code name from Intel website, and display the ESXi model, CPU model information at deploy_vm failure. Failures on 70GA, 71U1 with Intel Skylake, Ice Lake, and Cascade Lake CPU, print known issue.

failed_tasks.log
```
2024-03-02 11:00:48,002 | Failed at Play [1_deploy_vm] *******************************

2024-03-02 11:00:48,002 | TASK [1_deploy_vm][Check VMware Tools is running and collects guest OS fullname successfully] 
task path: /home/qiz/workspace/github/ansible-vsphere-gos-validation/common/vm_wait_guest_fullname.yml:46
fatal: [localhost]: FAILED! => {
    "assertion": "vm_guestinfo.instance.guest.toolsRunningStatus == \"guestToolsRunning\"",
    "changed": false,
    "evaluated_to": false,
    "msg": "It's timed out for VMware Tools collecting guest OS fullname in 5 seconds. Current VMware Tools running status is 'guestToolsNotRunning', and guest OS fullname is ''."
}
error message:
It's timed out for VMware Tools collecting guest OS fullname in 5 seconds. Current VMware Tools running status is 'guestToolsNotRunning', and guest OS fullname is ''.


2024-03-02 11:00:49,002 | TASK [1_deploy_vm][Display ESXi and CPU model information at deployment failure] 
task path: /home/qiz/workspace/github/ansible-vsphere-gos-validation/linux/deploy_vm/deploy_vm.yml:63
ok: [localhost] => {
    "msg": "VM deployment failed on server VMware, Inc. VMware7,1 with ESXi version 7.0.1 build 16850804 installed. The server's CPU model is Intel(R) Xeon(R) Gold 6330 CPU @ 2.00GHz (code name Ice Lake)"
}
error message:
VM deployment failed on server VMware, Inc. VMware7,1 with ESXi version 7.0.1 build 16850804 installed. The server's CPU model is Intel(R) Xeon(R) Gold 6330 CPU @ 2.00GHz (code name Ice Lake)
```

known_issues.log
```
2024-03-02 11:00:49,002 | Known Issue in Play [1_deploy_vm] **************************

2024-03-02 11:00:49,002 | TASK [1_deploy_vm][Known issue - Linux guest OS hung] ******
task path: /home/qiz/workspace/github/ansible-vsphere-gos-validation/linux/deploy_vm/deploy_vm.yml:73
[...]

PR #562, opened 2024-03-26
[Windows] Skip VBS test for guest OS with lower hardware version

123lzxm:
Using Intel CPUs for VBS requires vSphere 6.7 or later (HW14)
Using AMD CPUs for VBS equires vSphere 7.0 Update 2 or later (HW19)

### DIFF
--- a/common/esxi_get_model.yml
+++ b/common/esxi_get_model.yml
@@ -21,7 +21,7 @@
   ansible.builtin.set_fact:
     esxi_model_info: "{{ esxi_host_property_results.ansible_facts.summary.hardware.vendor }} {{ esxi_host_property_results.ansible_facts.summary.hardware.model }}"
     esxi_cpu_model_info: "{{ esxi_host_property_results.ansible_facts.summary.hardware.cpuModel }}"
-    esxi_cpu_vendor: "{{ esxi_host_property_results.ansible_facts.hardware.cpuPkg[0].vendor | default('Unknown') }}"
+    esxi_cpu_vendor: "{{ esxi_host_property_results.ansible_facts.hardware.cpuPkg[0].vendor | default('Unknown') | lower }}"
 
 - name: "Get ESXi server CPU code name"
   when: esxi_cpu_vendor == 'intel'
--- a/common/esxi_get_model.yml
+++ b/common/esxi_get_model.yml
@@ -24,7 +24,7 @@
     esxi_cpu_vendor: "{{ esxi_host_property_results.ansible_facts.hardware.cpuPkg[0].vendor | default('Unknown') }}"
 
 - name: "Get ESXi server CPU code name"
-  when: esxi_cpu_vendor | lower == 'intel'
+  when: esxi_cpu_vendor == 'intel'
   block:
     - name: "Get ESXi server CPU code name"
       ansible.builtin.script: "../tools/query_cpu.py -n '{{ cpu_name }}'"
--- a/windows/vbs_enable_disable/vbs_enable_disable.yml
+++ b/windows/vbs_enable_disable/vbs_enable_disable.yml
@@ -40,6 +40,18 @@
         - name: "Skip test case"
           include_tasks: ../../common/skip_test_case.yml
           vars:
+            skip_msg: >-
+              Skip test case due to using Intel CPUs for VBS requires vSphere 6.7 or later and using AMD CPUs for VBS
+              requires vSphere 7.0 Update 2 or later. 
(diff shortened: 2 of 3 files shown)


windows/vbs_enable_disable/vbs_enable_disable.yml
```diff
@@ -37,6 +37,18 @@
...
+        - name: "Skip test case"
+          include_tasks: ../../common/skip_test_case.yml
+          vars:
+            skip_msg: >-
+              Skip test case due to using Intel CPUs for VBS requires vSphere 6.7 or later and using AMD CPUs for VBS
+              requires vSphere 7.0 Update 2 or later. 
+              Please refer to https://docs.vmware.com/en/VMware-vSphere/8.0/vsphere-security/GUID-CE292D3F-D4AC-4607-B262-DE19CE6E9F6B.html
+          skip_reason: "Not Supported"
+          when: >-
+            (vm_hardware_version_num | int < 14 and esxi_cpu_model == "Intel") or
```
Tomorrow9, line 49: Qi added a parameter `esxi_cpu_vendor` in merge request #554

123lzxm: esxi_cpu_model_info is coming from "esxi_host_property_results.ansible_facts.summary.hardware.cpuModel", while esxi_cpu_vendor is coming from esxi_host_property_results.ansible_facts.hardware.cpuPkg[0].vendor. 

I think for this scenario, cpuModel is more precise.

Tomorrow9: If that info is already defined, then here you no need to define it again:
```
 - name: "Set ESXi CPU model"
          ansible.builtin.set_fact:
            esxi_cpu_model: |-
              {%- if 'AMD' in esxi_cpu_model_info -%}AMD
              {%- elif 'Intel' in esxi_cpu_model_info -%}Intel
              {%- else -%}Others
              {%- endif -%}
```
if you find that value is not accurate, then we need to change it instead of defining a new one.

123lzxm: Here does not create a new one, just use the value of esxi_cpu_model_info got in esxi_get_model.yml. 
When above code was added for VBS, there is only variable esxi_cpu_model_info, no esxi_cpu_vendor.  esxi_cpu_vendor was added later, not sure why https://github.com/vmware/ansible-vsphere-gos-validation/pull/554 does not use the already existing one.

123lzxm: As we talked offline, will use esxi_cpu_vendor here.

windows/vbs_enable_disable/vbs_enable_disable.yml
```diff
@@ -29,13 +29,17 @@
...
+          include_tasks: ../../common/skip_test_case.yml
+          vars:
+            skip_msg: >-
+              Skip test case due to using Intel CPUs for VBS requires vSphere 6.7 or later and using AMD CPUs for VBS requires 
+              vSphere 7.0 Update 2 or later. Current CPU is {{ esxi_cpu_vendor }} and hardware version is {{ vm_hardware_version_num }}.
+              Please refer to https://docs.vmware.com/en/VMware-vSphere/8.0/vsphere-security/GUID-CE292D3F-D4AC-4607-B262-DE19CE6E9F6B.html
+            skip_reason: "Not Supported"
+          when: >-
+            (vm_hardware_version_num | int < 14 and esxi_cpu_vendor | lower == "intel") or
+            (vm_hardware_version_num | int < 19 and esxi_cpu_vendor | lower == "amd")
```
keirazhang, line 42: `esxi_cpu_vendor`'s value is in lower case, so you can directly use it without `lower` filter. And I suggest you  use hardware version instead of vSphere version in skip message, because the limitation is coming from the hardware version actually. For example, if the vSphere version is 8.0.2 but VM's hardware version is 13, it should not work for VBS.

123lzxm: Sure, thanks.

common/esxi_get_model.yml
```diff
@@ -24,7 +24,7 @@
     esxi_cpu_vendor: "{{ esxi_host_property_results.ansible_facts.hardware.cpuPkg[0].vendor | default('Unknown') }}"
 
 - name: "Get ESXi server CPU code name"
-  when: esxi_cpu_vendor | lower == 'intel'
```
Tomorrow9, line 27: If 'lower' is removed from where it's called, then it need to be added when it's defined to make sure the value is indeed in lower case.

123lzxm: Add the filter to its definition.

Tomorrow9 (approved):
LGTM, thanks!

PR #567, opened 2024-04-01
[Linux] update get_system_firmware.yml to get system firmware for FreeBSD

keirazhang:
FreeBSD doesn't have `/sys/firmware/efi`, but can check `efirt` is loaded if system firmware is EFI. So, this fix is to set `guest_firmware` for FreeBSD.

Test passed:
```
+----------------------------------------------------------+
| GuestInfo Detailed Data   | architecture='X86'           |
|                           | bitness='64'                 |
|                           | distroName='FreeBSD'         |
|                           | familyName='FreeBSD'         |
|                           | kernelVersion='13.3-RELEASE' |
+----------------------------------------------------------+


Test Results (Total: 1, Passed: 1, Elapsed Time: 00:01:10)
+----------------------------------------------+
| ID | Name               | Status | Exec Time |
+----------------------------------------------+
|  1 | check_efi_firmware | Passed | 00:00:52  |
+----------------------------------------------+
```

### DIFF
--- a/linux/utils/get_system_firmware.yml
+++ b/linux/utils/get_system_firmware.yml
@@ -1,13 +1,13 @@
 # Copyright 2021-2024 VMware, Inc.
 # SPDX-License-Identifier: BSD-2-Clause
 #
-# Get system firmware
+# Get guest OS system firmware is EFI or not
 # Return:
-#   guest_firmware: EFI or BIOS
+#   guest_firmware_is_efi: return true if the system firmware is EFI, otherwise return false.
 #
-- name: "Initialize the fact of guest OS system firmware"
+- name: "Initialize the fact of guest OS system firmware is EFI or not"
   ansible.builtin.set_fact:
-    guest_firmware: ""
+    guest_firmware_is_efi: false
 
 - name: "Get firmware for Linux system"
   when: guest_os_ansible_system == "linux"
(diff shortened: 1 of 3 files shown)


linux/utils/get_system_firmware.yml
```diff
@@ -5,17 +5,31 @@
...
+  block:
+    - name: "Check 'efirt' module is loaded or not"
+      ansible.builtin.shell: "kldstat -m efirt"
+      delegate_to: "{{ vm_guest_ip }}"
+      ignore_errors: true
+      register: kldstat_efirt_result
+
+    - name: "Set fact of guest OS system firmware"
+      ansible.builtin.set_fact:
+        guest_firmware: "{{ 'EFI' if kldstat_efirt_result.rc is defined and kldstat_efirt_result.rc == 0 else 'BIOS' }}"
```
Tomorrow9, line 35: How about above task failed situation?

keirazhang: If task failed, it will be set to "BIOS".

Tomorrow9: But you don't know the reason causing that task failure and set it to BIOS directly.

keirazhang: Any suggestion?

Tomorrow9: I think getting firmware type tasks can fail as it is, so that we can get attention on things changed or things broken, just set it with the value when tasks succeed.

keirazhang: Changed the variable name to `guest_firmware_is_efi` to represent whether guest firmware is efi or not.

PR #571, opened 2024-04-07
[Linux] Update packagelist_file in VMware Photon OS kickstart

keirazhang:
The package list file "packages_ova.json" was removed from VMware Photon OS 5.0 daily build. This fix replaced it with "packages_minimal.json".


123lzxm (approved):
LGTM, thanks.

PR #574, opened 2024-04-18
[Windows] Skip MDAG test for build greater than 26052

123lzxm:
From build 26052, MDAG is deprecated. Skip MDAG test from the build.


Tomorrow9 (approved):
LGTM， thanks

PR #576, opened 2024-04-23
Update guestinfo_detailed_data on newer vSphere with latest VMware Tools

keirazhang:
VM with latest VMware Tools on newer vSphere (looks like 8.0U2) post the detailed data to guest info instead of VM extra config. This fix is to set `guestinfo_detailed_data` with  `vm_guestinfo.instance.guest.guestDetailedData` if it is not empty.


keirazhang:
It happened once that vm_extra_config doesn't have detailed data, but rerun didn't reproduce it. Anyway, `vm_guestinfo.instance.guest.guestDetailedData` exists on newer vSphere.

123lzxm (approved):
LGTM, thanks.

PR #577, opened 2024-04-28
[FreeBSD 13 32bit] Fix autoinstall failure due to no kde5 package from quarterly repo

ZouYuhua:
Issue:
[FreeBSD 13 32bit] autoinstall is failed due to not to get kde5 package from repo

Result:
+----------------------------------------------------------+
| GuestInfo Detailed Data   | architecture='X86'           |
|                           | bitness='32'                 |
|                           | distroName='FreeBSD'         |
|                           | familyName='FreeBSD'         |
|                           | kernelVersion='13.3-RELEASE' |
+----------------------------------------------------------+

Test Results (Total: 1, Passed: 1, Elapsed Time: 00:45:45)
+-------------------------------------------------------+
| ID | Name                        | Status | Exec Time |
+-------------------------------------------------------+
|  1 | deploy_vm_bios_nvme_vmxnet3 | Passed | 00:45:26  |
+-------------------------------------------------------+

### DIFF
--- a/autoinstall/FreeBSD/13/installerconfig
+++ b/autoinstall/FreeBSD/13/installerconfig
@@ -81,8 +81,8 @@ rm -rf /usr/local/etc/pkg/repos/FreeBSD_install_cdrom.conf
 
 # Modify the offical repo to latest
 mkdir -p /usr/local/etc/pkg/repos
-cat >/usr/local/etc/pkg/repos/FreeBSD.conf <<EOF
-FreeBSD: {
+cat >/usr/local/etc/pkg/repos/FreeBSD_latest.conf <<EOF
+FreeBSD_latest: {
     url: "pkg+http://pkg.FreeBSD.org/\${ABI}/latest",
     mirror_type: "srv",
     signature_type: "fingerprints",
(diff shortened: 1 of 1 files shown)


autoinstall/FreeBSD/13/installerconfig
```diff
@@ -78,6 +78,21 @@ done
...
+cat >/usr/local/etc/pkg/repos/FreeBSD.conf <<EOF
+FreeBSD: {
+    url: "pkg+http://pkg.FreeBSD.org/\${ABI}/latest",
+    mirror_type: "srv",
+    signature_type: "fingerprints",
+    fingerprints: "/usr/share/keys/pkg",
+    enabled: yes
+}
+EOF
+{% endif %}
```
keirazhang, line 94: What if keep both of quarterly repo and latest repo?  And I think this change can be applied to both 32bit and 64bit FreeBSD. So that we can install same version of packages on 32bit and 64bit FreeBSD

ZouYuhua: > What if keep both of quarterly repo and latest repo?
The Note in file /etc/pkg/FreeBSD.conf :
    To disable this repository. Instead of modifying or removing this file. create a /usr/local/etc/pkg/repos/FreeBSD.conf file.

> And I think this change can be applied to both 32bit and 64bit FreeBSD. So that we can install same version of packages on 32bit and 64bit FreeBSD

OK

ZouYuhua:
@keirazhang 

After the code review:
1. Pass the testing for freebsd 13.2 32bit:
image


2. Pass the testing for freebsd 13.2 64bit:
image

keirazhang (approved):
LGTM. Thanks!

PR #592, opened 2024-06-19
Force to install latest required packages on Photon OS

keirazhang:
When deploying VM from Photon OS 4.0 OVA, it can't install required packages `python3-rpm` at cloud-init config because the expired 1024 GPG key, which would led to getting installed packages failed or installing packages failed in the VM because of:
```
task path: /home/worker/workspace/Ansible_Regression_Photon_4.0_OVA/ansible-vsphere-gos-validation/linux/utils/get_installed_packages.yml:12
fatal: [localhost -> 10.185.15.85]: FAILED! => {
    "changed": false,
    "invocation": {
        "module_args": {
            "manager": [
                "auto"
            ],
            "strategy": "first"
        }
    },
    "msg": "Could not detect a supported package manager from the following list: ['pkg_info', 'apk', 'portage', 'pkg', 'apt', 'pacman', 'rpm'], or the required Python library is not installed. Check warnings for details.",
    "warnings": [
        "Found \"rpm\" but Failed to import the required Python library (rpm) on photon-ova-20240614115745's Python /usr/bin/python3.10. Please read the module documentation and install it in the appropriate location. If the required library is installed, but Ansible is using the wrong Python interpreter, please consult the documentation on ansible_python_interpreter",
        "Platform linux on host localhost is using the discovered Python interpreter at /usr/bin/python3.10, but future installation of another Python interpreter could change the meaning of that path. See https://docs.ansible.com/ansible-core/2.15/reference_appendices/interpreter_discovery.html for more information."
    ]
}
error message:
Could not detect a supported package manager from the following list: ['pkg_info', 'apk', 'portage', 'pkg', 'apt', 'pacman', 'rpm'], or the required Python library is not installed. Check warnings for details.
```

This fix forces Photon OS VM to install latest required packages, so that it can update GPG key firstly, then install those packages.

### DIFF
--- a/linux/setup/test_setup.yml
+++ b/linux/setup/test_setup.yml
@@ -38,6 +38,10 @@
   include_tasks: ../utils/get_linux_system_info.yml
   when: guest_os_system_info_retrieved is undefined or not guest_os_system_info_retrieved
 
+- name: "Take a base snapshot if it does not exist"
+  include_tasks: create_base_snapshot.yml
+  when: not (base_snapshot_exists | bool)
+
 - name: "Get VMware Tools version and build"
   include_tasks: ../utils/get_guest_ovt_version_build.yml
   when:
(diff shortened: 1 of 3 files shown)


123lzxm (approved):
LGTM, thanks.

linux/setup/test_setup.yml
```diff
@@ -38,6 +38,10 @@
   include_tasks: ../utils/get_linux_system_info.yml
   when: guest_os_system_info_retrieved is undefined or not guest_os_system_info_retrieved
 
+- name: "Take a base snapshot if it does not exist"
```
Tomorrow9, line 41: Why move taking base snapshot above here, VMware Tools status is not checked already running.

keirazhang: line 25 gets vmtools status. `get_guest_ovt_version_build.yml` will call `get_installed_packages.yml` to check whether ovt package is installed. If Photon OS VM doesn't have python3-rpm installed, `get_installed_packages.yml` will fail due to lacking of dependency. So move `create_base_snapshot.yml` before `get_guest_ovt_version_build.yml` to force Photon OS to install all required packages, no matter whether it's installed or not.

Tomorrow9 (approved):
LGTM, thanks

PR #598, opened 2024-06-25
[Linux] Check guest has GUI from display-manager service or running Xorg or Xwayland process

keirazhang:
Xorg server has been deprecated in latest RHEL and Ubuntu releases, which has been replaced by Wayland. 
In our tests, checking Xorg installed or not is not precise for determine whether the guest OS has GUI. For example, the guest OS may install Xorg but has no desktop environment. 

For FreeBSD system, the display manager may relies on Xorg or Xwayland. So if there is running process of Xorg or Xwayland, it means the OS has desktop environment. 

For Linux system, we can also check running process of Xorg or Xwayland, but Linux guest OS coverage is more than FreeBSD, the same command couldn't work on all Linux system. As we also need to get Linux system display manager, so this fix directly check whether display-manager is running. If it is running, that means the OS has desktop environment.

### DIFF
--- a/linux/utils/check_guest_os_gui.yml
+++ b/linux/utils/check_guest_os_gui.yml
@@ -8,64 +8,67 @@
     guest_os_with_gui: false
     guest_os_display_manager: ""
 
-# Set default guest_os_with_gui by checking Xorg exists or not
-- name: "Check Xorg exists or not"
-  ansible.builtin.shell: "type Xorg"
-  ignore_errors: true
-  delegate_to: "{{ vm_guest_ip }}"
-  register: xorg_result
-
-- name: "Set facts of guest OS desktop environment"
-  when:
-    - xorg_result is defined
-    - xorg_result.rc is defined
-    - xorg_result.rc == 0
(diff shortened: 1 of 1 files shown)


linux/utils/check_guest_os_gui.yml
```diff
@@ -8,64 +8,67 @@
...
-    - xorg_result is defined
-    - xorg_result.rc is defined
-    - xorg_result.rc == 0
+- name: "Check Linux guest OS has desktop environment or not"
+  when: guest_os_ansible_system == "linux"
   block:
-    - name: "Set fact of guest OS having desktop environment"
+    - name: "Get guest OS display manager service detail"
+      ansible.builtin.shell: "systemctl status display-manager.service"
+      register: "display_manager_status"
```
Tomorrow9, line 16: Not need quotes here.

keirazhang: Fixed. Thanks!

Tomorrow9 (approved):
LGTM, thanks.

PR #599, opened 2024-06-28
[Linux] Skip LSILogic and LSILogicSAS testing for RHCOS 4.16.0 or later

keirazhang:
Since RHCOS 4.16.0, LSILogic and LSILogicSAS  controllers are not supported. Skip the two vHBA test cases.


Tomorrow9 (approved):
LGTM, thanks.

PR #608, opened 2024-07-10
[Linux] Update GOSC test scope and handle known issues with workaround

keirazhang:
1. Amazon Linux 2 and 2023 is not officially supported by vSphere. So here excluded it from GOSC supporting list. When it is officially supported, we can then enable its GOSC testing and ask for GOSC team to fix its failures.
2. For RHEL family OS like AlmaLinux, Rocky Linux, Oracle Linux and CentOS, their GOSC support status is as same as RHEL since release 9.0. To reduce redundant settings in `linux/guest_customization/gosc_support_matrix.yml`, these OS will follow RHEL's GOSC support matrix.
3. MIRACLE LINUX is a bit different. Its GOSC was supported since 8.0.3 release. So it has different support matrix.
4. VMware KB links will be redirected to Broadcom KB links. So this fix also replaced VMware KB links with Broadcom links. For VMware KB links in other test cases, I will create a new PR to change them.
5. For known issues with workaround, they are handled in `linux/guest_customization/handle_gosc_known_issues.yml`.  Currently 3 known issues were handled:
1. From CentOS 10 and RHEL 10 , they don't have `/etc/sysconfig/network-scripts`, which will lead to GOSC failure on vSphere 8.0.x. In vSphere 8.0.3, an workaround fix was provided, which is to manually create this directory. Then GOSC scripts can create network config file under it. 
2. From Ubuntu 23.10 release, it doesn't have hwclock by default, which leads to perl GOSC fail on 8.0.2 and earlier releases. The workaround is to install the package. 8.0.3 has fixed it, so it won't need to install the package since 8.0.3.
3. Oracle Linux 8.7 & 9.1 built-in cloud-init has bugs, which led to cloud-init GOSC failure. The workaround is to upgrade cloud-init.


linux/guest_customization/handle_gosc_known_issues.yml
```diff
@@ -0,0 +1,93 @@
...
+                - known_issue
+
+    # Perl GOSC fails on Ubuntu 23.10 or later versions due to hwclock not built-in
+    - name: "Handle known issue for {{ vm_guest_os_distribution }} on vCenter Server {{ vcenter_version }}"
+      when:
+        - vcenter_version is version('8.0.3', '<')
+        - guest_os_ansible_distribution == "Ubuntu"
+        - guest_os_ansible_distribution_ver is version('23.10', '>=')
+      block:
+        - name: "Uninstall current cloud-init"
```
Tomorrow9, line 45: It's 'Install' instead of 'Uninstall'?

linux/guest_customization/handle_gosc_known_issues.yml
```diff
@@ -0,0 +1,93 @@
...
+      include_tasks: ../utils/install_uninstall_package.yml
+      vars:
+        package_list: ["cloud-init"]
+        package_state: "latest"
+
+    - name: "Known issue - failure of cloud-init GOSC in Oracle Linux 8.7 and 9.1"
+      ansible.builtin.debug:
+        msg:
+          - "Cloud-init guest customization would fail because cloud-init-local.service couldn't start when using cloud-init 22.1-5.0.1."
+          - "The issue has been resolved in cloud-init 22.1-6.0.1. In this case cloud-init is upgraded to latest version as a workaround."
```
Tomorrow9, line 90: Like statements above, here can be changed:
In this case cloud-init is upgraded to latest version as a workaround. -> Upgrade cloud-init to the latest version as a workaround.

linux/guest_customization/check_gosc_support_status.yml
```diff
@@ -89,25 +90,44 @@
...
+            - guest_os_ansible_distribution_major_ver | int >= 9
+
+        - name: "Set fact of Perl GOSC matrix for {{ vm_guest_os_distribution }}"
+          ansible.builtin.set_fact:
+            guest_gosc_suport_matrix: "{{ gosc_support_matrix[guest_os_ansible_distribution] }}"
+          when:
+            - (guest_os_family != "RedHat" or
+               guest_os_ansible_distribution == 'MIRACLE' or
+               guest_os_ansible_distribution_major_ver | int < 9)
+            - gosc_support_matrix[guest_os_ansible_distribution] is defined
```
Tomorrow9, line 123: Can this line also be set as the '| default({})' as above?

linux/guest_customization/check_gosc_support_status.yml
```diff
@@ -119,35 +139,38 @@
...
                     - name: "Get {{ guest_os_ansible_distribution }} {{ guest_os_ansible_distribution_ver }} Perl GOSC supported open-vm-tools version"
                       ansible.builtin.set_fact:
-                        gosc_matrix_vmtools_version: "{{ gosc_support_matrix[guest_os_ansible_distribution][gosc_matrix_first_os_version]['vmtools'] }}"
+                        gosc_matrix_vmtools_version: "{{ guest_gosc_suport_matrix[gosc_matrix_first_os_version]['vmtools'] }}"
 
                     - name: "Get {{ guest_os_ansible_distribution }} {{ guest_os_ansible_distribution_ver }} Perl GOSC support status with open-vm-tools {{ vmtools_version }}"
                       ansible.builtin.set_fact:
                         gosc_is_supported: false
-                      when: vmtools_version is version(gosc_support_matrix[guest_os_ansible_distribution][gosc_matrix_first_os_version]['vmtools'], '<')
+                      when: vmtools_version is version(guest_gosc_suport_matrix[gosc_matrix_first_os_version]['vmtools'], '<')
```
Tomorrow9, line 174: `gosc_is_supported | bool` can also be simplified.

linux/guest_customization/handle_gosc_known_issues.yml
```diff
@@ -0,0 +1,97 @@
...
+                - known_issue
+
+    # Perl GOSC fails on Ubuntu 23.10 or later versions due to hwclock not built-in
+    - name: "Handle known issue for {{ vm_guest_os_distribution }} on vCenter Server {{ vcenter_version }}"
+      when:
+        - vcenter_version is version('8.0.3', '<')
+        - guest_os_ansible_distribution == "Ubuntu"
+        - guest_os_ansible_distribution_ver is version('23.10', '>=')
+      block:
+        - name: "Install 'hwclock' package 'util-linux-extra'"
```
Tomorrow9, line 49: It's better to check if 'hwclock' exists firstly as above known issue.

linux/guest_customization/check_gosc_support_status.yml
keirazhang: Line 105 has already initialized the value of `guest_gosc_suport_matrix` to `{}`.  Here will only change it when `gosc_support_matrix[guest_os_ansible_distribution]` is defined in 'gosc_support_matrix.yml' for Miracle Linux and Linux other than RedHat family.

The reason to use `default({})` in above set_fact is mainly for OS like CentOS and Oracle Linux, which have completely same support matrix with RHEL but don't have their own definitions. Using `gosc_support_matrix[guest_os_ansible_distribution] | default({}` to avoid undefined error.

linux/guest_customization/check_gosc_support_status.yml
keirazhang: Fixed. Thanks!

linux/guest_customization/check_gosc_support_status.yml
keirazhang: Simplified setting guest_gosc_suport_matrix into 2 `set_facts` tasks instead of 3.

keirazhang: Just realized that it can't be simplified as `guest_gosc_suport_matrix` cannot use `gosc_support_matrix` in the same `set_fact` task. So I reverted the change

linux/guest_customization/check_gosc_support_status.yml
```diff
@@ -105,26 +105,20 @@
         - name: "Set facts of Perl GOSC support matrix"
           ansible.builtin.set_fact:
             gosc_support_matrix: "{{ lookup('file', 'gosc_support_matrix.yml') | from_yaml }}"
-            guest_gosc_suport_matrix: {}
+
+        - name: "Set facts of Perl GOSC support matrix for {{ guest_os_ansible_distribution }}"
```
Tomorrow9, line 109: facts -> fact

Tomorrow9 (approved):
LGTM, thanks!

PR #614, opened 2024-07-19
[Windows] windows_online_updates_install result is 'Not Applicable' when there are updates installed

Tomorrow9:
In the situation that "found updates count != installed updates count", while the reason is some updates are not applicable to the OS and the task result is not 'failed', so this is acceptable result.

### DIFF
--- a/windows/utils/win_install_online_updates.yml
+++ b/windows/utils/win_install_online_updates.yml
@@ -107,8 +107,7 @@
           block:
             - name: "Set fact of failure messages can not be ignored"
               ansible.builtin.set_fact:
-                win_updates_failure_msgs: >
-                  {{ win_updates_failure_msgs | difference(win_updates_failure_msgs | select('search', item)) }}
+                win_updates_failure_msgs: "{{ win_updates_failure_msgs | reject('search', item) }}"
               loop: "{{ win_updates_ignore_kws }}"
             - name: "Set fact of Windows online updates install status"
               ansible.builtin.set_fact:
--- a/windows/utils/win_install_online_updates.yml
+++ b/windows/utils/win_install_online_updates.yml
@@ -26,8 +26,8 @@
 #        For example, 'failure_msg' is "Operation was not performed because there are
 #        no applicable updates (WU_E_NOT_APPLICABLE 0x80240017)".
 #     Other cases, it will be set to 'false'.
-#   win_online_updates_installed: If set to 'true', there are Windows online updates
-#     installed in guest OS, or it will be 'false'.
+#   win_online_updates_installed: If set to 'true', there are at least one Windows online
+#    updates installed in guest OS, or it will be 'false'.
 #
 - name: "Set fact and initialize Windows online updates install parameters"
   ansible.builtin.set_fact:
@@ -96,14 +96,20 @@
         - name: "Set fact of failure messages of not installed updates"
           ansible.builtin.set_fact:
             win_updates_failure_msgs: "{{ win_updates_not_installed | map(attribute='failure_msg') }}"
+        - name: "Set fact of the lengths of failure messages and not installed updates lists"
+          ansible.builtin.set_fact:
+            win_updates_failure_msgs_len: "{{ win_updates_failure_msgs | length }}"
+            win_updates_not_installed_len: "{{ win_updates_not_installed | length }}"
 
-        - name: "No failure message for update"
+        - name: "No failure messages for not installed updates"
           ansible.builtin.debug:
-            msg: "No 'failure_msg' info for not installed updates, so will not ignore these not installed updates."
-          when: win_updates_failure_msgs | length != win_updates_not_installed | length
+            msg: >-
+              {{ win_updates_not_installed_len - win_updates_failure_msgs_len }} updates
+              are not installed and without failure messages, which can not be ignored.
+          when: win_updates_failure_msgs_len != win_updates_not_installed_len
 
         - name: "Get failure messages for all not installed updates"
-          when: win_updates_failure_msgs | length == win_updates_not_installed | length
+          when: win_updates_failure_msgs_len == win_updates_not_installed_len
           block:
             - name: "Set fact of failure messages can not be ignored"
               ansible.builtin.set_fact:
--- a/windows/utils/win_install_online_updates.yml
+++ b/windows/utils/win_install_online_updates.yml
@@ -96,14 +96,24 @@
         - name: "Set fact of failure messages of not installed updates"
           ansible.builtin.set_fact:
             win_updates_failure_msgs: "{{ win_updates_not_installed | map(attribute='failure_msg') }}"
-        - name: "Set fact of failure messages can not be ignored"
-          ansible.builtin.set_fact:
-            win_updates_failure_msgs: "{{ win_updates_failure_msgs | difference(win_updates_failure_msgs | select('search', item)) }}"
-          loop: "{{ win_updates_ignore_kws }}"
-        - name: "Set fact of Windows online updates install status"
-          ansible.builtin.set_fact:
-            win_online_updates_succeed: true
-          when: win_updates_failure_msgs | length == 0
+
+        - name: "No failure message for update"
+          ansible.builtin.debug:
+            msg: "No 'failure_msg' info for not installed updates, so will not ignore these not installed updates."
+          when: win_updates_failure_msgs | length != win_updates_not_installed | length
+
+        - name: "Get failure messages for all not installed updates"
+          when: win_updates_failure_msgs | length == win_updates_not_installed | length
+          block:
+            - name: "Set fact of failure messages can not be ignored"
+              ansible.builtin.set_fact:
+                win_updates_failure_msgs: >
+                  {{ win_updates_failure_msgs | difference(win_updates_failure_msgs | select('search', item)) }}
+              loop: "{{ win_updates_ignore_kws }}"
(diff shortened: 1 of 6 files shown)


windows/utils/win_install_online_updates.yml
```diff
@@ -46,10 +46,12 @@
   ansible.builtin.set_fact:
     win_online_updates_installed: true
   when:
+    - win_updates_install_result.failed is defined
+    - not win_updates_install_result.failed
```
123lzxm, line 50: Will win_updates_install_result.failed be set to true if there's indeed some updates installation failure, not because of not applicable to OS?

Tomorrow9: Don't know, we can handle all other situations when we meet.

Tomorrow9: If `win_updates_install_result.failed` is true, then there might be something wrong during the installation, either task failed when not ignore errors, or `win_online_updates_installed` is not true.

123lzxm: I suggest we can add two vars to identify different scenarios. One is for the scenario there's at lease one update installed. The other is for the number of installed updates is equal to the number of found updates.

windows/windows_update_install/install_online_updates.yml
```diff
@@ -3,7 +3,7 @@
 ---
 - name: "Initialize Windows online updates install status"
   ansible.builtin.set_fact:
-    win_online_updates_installed: false
+    win_online_updates_succeed: false
```
keirazhang, line 6: This task can be removed since this var is defined in `win_install_online_updates.yml`, and it is used after `win_install_online_updates.yml`.

windows/utils/win_install_online_updates.yml
Tomorrow9: Changed to return "win_online_updates_succeed" and "win_online_updates_installed". Thanks.

windows/utils/win_install_online_updates.yml
```diff
@@ -39,19 +51,62 @@
...
+            win_updates_not_installed: "{{ win_updates_install_result.updates.values() | selectattr('installed', 'equalto', false) }}"
+        - name: "Display the not installed updates list"
+          ansible.builtin.debug: var=win_updates_not_installed
+
+        - name: "Set fact of failure messages of not installed updates"
+          ansible.builtin.set_fact:
+            win_updates_failure_msgs: "{{ win_updates_not_installed | map(attribute='failure_msg') }}"
+        - name: "Set fact of failure messages can not be ignored"
+          ansible.builtin.set_fact:
+            win_updates_failure_msgs: "{{ win_updates_failure_msgs | difference(win_updates_failure_msgs | select('search', item)) }}"
```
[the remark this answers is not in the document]
Tomorrow9: Yeah, `reject` is what I need here.
It's case sensitive since we only can ignore the exact what we met can be ignored. Thanks.

windows/windows_update_install/install_online_updates.yml
Tomorrow9: Removed, thanks.

windows/utils/win_install_online_updates.yml
```diff
@@ -16,13 +16,25 @@
...
+#     2. there are Windows updates found, and the installed updates count is equal to
+#        the found count,
+#     3. there are Windows updates found, and the installed updates count is not equal
+#        to the found count, but the failure messages of the not installed ones contain
+#        the keywords representing failure can be ignored.
+#        For example, 'failure_msg' is "Operation was not performed because there are
+#        no applicable updates (WU_E_NOT_APPLICABLE 0x80240017)".
+#     Other cases, it will be set to 'false'.
+#   win_online_updates_installed: If set to 'true', there are Windows online updates
+#     installed in guest OS, or it will be 'false'.
```
123lzxm, line 30: Maybe it's more clear. 
"there are Windows online updates installed in guest OS" ---> "there are **at least one** Windows online updates installed in guest OS".

windows/utils/win_install_online_updates.yml
```diff
@@ -39,17 +51,71 @@
...
+          ansible.builtin.set_fact:
+            win_updates_not_installed: "{{ win_updates_install_result.updates.values() | selectattr('installed', 'equalto', false) }}"
+        - name: "Display the not installed updates list"
+          ansible.builtin.debug: var=win_updates_not_installed
+
+        - name: "Set fact of failure messages of not installed updates"
+          ansible.builtin.set_fact:
+            win_updates_failure_msgs: "{{ win_updates_not_installed | map(attribute='failure_msg') }}"
+
+        - name: "No failure message for update"
```
123lzxm, line 100: I think the task is not needed. It may make user a litter confused.

Or just keep the first part of the message. The last part seems a little complex to understand.

windows/windows_online_updates_install/install_windows_online_updates.yml
```diff
@@ -44,15 +44,28 @@
     skip_reason: "Not Applicable"
   when:
     - not win_check_is_updated
+    - win_online_updates_succeed
     - not win_online_updates_installed
 
-# Fail test case when no update installed while expect guest OS to be updated
-- name: "No Windows updates installed in guest OS"
```
123lzxm, line 50: If the expected build number is not set and no need to check the OS build number, I think we can just print the not installed windows updates list, no need to fail the test case.

windows/utils/win_install_online_updates.yml
keirazhang: Maybe the debug message can be changed to:
```
 {{ win_updates_not_installed | length - win_updates_failure_msgs | length }} updates failed to be installed without failure messages, which can not be ignored.
```
or something like this?

windows/utils/win_install_online_updates.yml
```diff
@@ -39,17 +51,71 @@
...
+          ansible.builtin.debug:
+            msg: "No 'failure_msg' info for not installed updates, so will not ignore these not installed updates."
+          when: win_updates_failure_msgs | length != win_updates_not_installed | length
+
+        - name: "Get failure messages for all not installed updates"
+          when: win_updates_failure_msgs | length == win_updates_not_installed | length
+          block:
+            - name: "Set fact of failure messages can not be ignored"
+              ansible.builtin.set_fact:
+                win_updates_failure_msgs: "{{ win_updates_failure_msgs | reject('search', item) }}"
```
keirazhang, line 110: If the win_updates_failure_msgs is changed, the value of `win_updates_failure_msgs | length` at line 106 will be changed, and then it will quit from this block, and never update `win_online_updates_succeed`.

You can add 2 vars to save the value of `win_updates_failure_msgs | length` and `win_updates_not_installed | length`, which can avoid this situation.

windows/utils/win_install_online_updates.yml
```diff
@@ -16,13 +16,25 @@
...
+#     Other cases, it will be set to 'false'.
+#   win_online_updates_installed: If set to 'true', there are Windows online updates
+#     installed in guest OS, or it will be 'false'.
 #
 - name: "Set fact and initialize Windows online updates install parameters"
   ansible.builtin.set_fact:
     win_updates_log_file: "{{ win_updates_log_file | default('C:\\win_updates_log.txt') }}"
+    win_online_updates_succeed: false
     win_online_updates_installed: false
+    win_updates_ignore_kws: ["no applicable"]
```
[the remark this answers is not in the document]
Tomorrow9: We don't have many samples that can say the such keyword is in the failure message definitely, so both are ok here for this single situation we met. We can update it when we have other samples of failure messages. Thanks.

windows/utils/win_install_online_updates.yml
Tomorrow9: Updated, thanks.

windows/utils/win_install_online_updates.yml
Tomorrow9: Updated, thanks.

windows/windows_online_updates_install/install_windows_online_updates.yml
Tomorrow9: Will not check build number after updates install when `windows_updated_build_num` is not set, but updates install should succeed.

windows/utils/win_install_online_updates.yml
Tomorrow9: Yeah, right, thanks for catching this.

keirazhang (dismissed):
LGTM. Thanks!

windows/utils/win_install_online_updates.yml
```diff
@@ -39,17 +51,77 @@
...
+
+        - name: "No failure messages for not installed updates"
+          ansible.builtin.debug:
+            msg: >-
+              {{ win_updates_not_installed_len - win_updates_failure_msgs_len }} updates
+              are not installed and without failure messages, which can not be ignored.
+          when: win_updates_failure_msgs_len != win_updates_not_installed_len
+
+        - name: "Get failure messages for all not installed updates"
+          when: win_updates_failure_msgs_len == win_updates_not_installed_len
```
123lzxm, line 112: Will it be win_updates_failure_msgs_len !=0? I.e. as long as there is failure message, should check if it can be ignored.

123lzxm: Ignore above comment.

123lzxm (dismissed):
LGTM, thanks.

keirazhang (approved):
LGTM. Thanks!

PR #618, opened 2024-08-06
Add mainline kernel upgarde support for ubuntu

linhuali:
- Add upgrade_kernel_version into var/test.yml for testing kernel upgrade only in ubuntu
- Testing done with ubuntu 24.04 and mainline kernel 6.11-rc2
```
ok: [localhost] => {
    "ansible_facts": {
        "upgrade_kernel_version": "6.11-rc2"
    },
    "changed": false
}
...
ok: [localhost] => {
    "ansible_facts": {
        "kernel_before_upgrade": "6.8.0-39-generic",
        "kernel_is_upgraded": false
    },
    "changed": false
}
...
ok: [localhost] => {
    "ansible_facts": {
        "kernel_after_upgrade": "6.11.0-061100rc2-generic"
    },
    "changed": false
}
```

### DIFF
--- a/linux/deploy_vm/deploy_vm_from_iso.yml
+++ b/linux/deploy_vm/deploy_vm_from_iso.yml
@@ -42,14 +42,9 @@
 - name: "Prepare for Ubuntu installation"
   when: guest_id is match('ubuntu.*')
   block:
-    - name: "Check upgrade_kernel_version when upgrade_kernel_version is defined"
+    - name: "Check and get upgrade_kernel_version when upgrade_kernel_version is defined"
+      include_tasks: ubuntu/get_mainline_kernel.yml
       when: upgrade_kernel_version is defined and upgrade_kernel_version
-      block:
-        - name: "Set the base url for kernel version check"
-          ansible.builtin.set_fact:
-            mainline_kernel_url: "https://kernel.ubuntu.com/mainline"
-        - name: "Check the kernel version exists"
-          include_tasks: ubuntu/get_mainline_kernel.yml
     - name: "Prepare for Ubuntu installation"
       include_tasks: ubuntu/prepare_ubuntu_iso_install.yml
 
@@ -312,7 +307,7 @@
         wait_service_status: "running"
 
 - name: "Upgrade kernel for {{ guest_os_ansible_distribution }} {{ guest_os_ansible_distribution_ver }}"
-  include_tasks: upgrade_vm_kernel.yml
+  include_tasks: upgrade_linux_kernel.yml
   when: >
     ((upgrade_kernel_version is defined and upgrade_kernel_version and 
     guest_os_ansible_distribution == "Ubuntu") or
--- a/linux/deploy_vm/deploy_vm_from_iso.yml
+++ b/linux/deploy_vm/deploy_vm_from_iso.yml
@@ -40,8 +40,18 @@
   include_tasks: ../../common/get_iso_file_list.yml
 
 - name: "Prepare for Ubuntu installation"
-  include_tasks: ubuntu/prepare_ubuntu_iso_install.yml
   when: guest_id is match('ubuntu.*')
+  block:
+    - name: "Check upgrade_kernel_version when upgrade_kernel_version is defined"
+      when: upgrade_kernel_version is defined and upgrade_kernel_version
+      block:
+        - name: "Set the base url for kernel version check"
+          ansible.builtin.set_fact:
+            mainline_kernel_url: "https://kernel.ubuntu.com/mainline"
+        - name: "Check the kernel version exists"
+          include_tasks: ubuntu/get_mainline_kernel.yml
+    - name: "Prepare for Ubuntu installation"
+      include_tasks: ubuntu/prepare_ubuntu_iso_install.yml
 
 - name: "Set default unattend install conf file"
   when: unattend_install_conf is undefined or not unattend_install_conf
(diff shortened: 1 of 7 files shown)


linux/deploy_vm/deploy_vm_from_iso.yml
```diff
@@ -40,8 +40,13 @@
   include_tasks: ../../common/get_iso_file_list.yml
 
 - name: "Prepare for Ubuntu installation"
-  include_tasks: ubuntu/prepare_ubuntu_iso_install.yml
   when: guest_id is match('ubuntu.*')
+  block:
+    - name: "Check and get upgrade_kernel_version when upgrade_kernel_version is defined"
+      include_tasks: ubuntu/get_mainline_kernel.yml
+      when: upgrade_kernel_version is defined and upgrade_kernel_version
```
keirazhang, line 47: Please set default value for upgrade_kernel_version to '' at the beginning of this file if it is not defined. Then you can remove all `upgrade_kernel_version is defined` checking, and only check its value in the following tasks.

linux/deploy_vm/ubuntu/get_mainline_kernel.yml
```diff
@@ -0,0 +1,45 @@
+# Copyright 2022-2024 VMware, Inc.
```
keirazhang, line 1: Copyright should be 2024

linux/deploy_vm/ubuntu/get_mainline_kernel.yml
```diff
@@ -0,0 +1,45 @@
...
+- name: "Set the base url for kernel version check"
+  ansible.builtin.set_fact:
+    mainline_kernel_url: "https://kernel.ubuntu.com/mainline"
+
+- name: "Check the kernel version exists"
+  ansible.builtin.uri:
+    url: "{{ mainline_kernel_url }}/v{{ upgrade_kernel_version }}/amd64"
+    return_content: true
+  register: check_kernel_result
+  ignore_errors: yes
```
keirazhang, line 13: Please replace "yes" with "true". If this task is just to check whether the url exists, `return_content: true` can be removed since you didn't use its content at all.

linux/deploy_vm/ubuntu/get_mainline_kernel.yml
```diff
@@ -0,0 +1,45 @@
+# Copyright 2022-2024 VMware, Inc.
+# SPDX-License-Identifier: BSD-2-Clause
+---
+- name: "Set the base url for kernel version check"
+  ansible.builtin.set_fact:
+    mainline_kernel_url: "https://kernel.ubuntu.com/mainline"
+
```
keirazhang, line 7: Its better to add a assert task to check user's setting about `upgrade_kernel_version`, it should be either `latest` or a valid version. Since a version is always carried with a 'v' prefix, I suggest you making the rule that user need to input version like`vX.Y.Z` not `X.Y.Z` etc

linux/deploy_vm/ubuntu/get_mainline_kernel.yml
```diff
@@ -0,0 +1,45 @@
...
+- name: "Get the latest kernel version"
+  when: >
+    (upgrade_kernel_version | lower == 'latest' or
+     upgrade_kernel_version is not match('[0-9]+.[0-9a-z.-]+') or
+     check_kernel_result.status is undefined or check_kernel_result.status != 200)
+  block:
+    - name: "Get the latest kernel with amd64"
+      ansible.builtin.shell: |
+        versions=`curl -s '{{ mainline_kernel_url }}/?C=N;O=D' | sed 's/>/\\n/g' |grep -Po '(?<=v)[0-9]+.[0-9a-z.-]+'|head -10`
```
keirazhang, line 24: Please make sure `|` has spaces before and after it. Why did this command not work on my Ubuntu 22.04.4 VM?

linux/deploy_vm/upgrade_linux_kernel.yml
```diff
@@ -0,0 +1,72 @@
...
+    - name: "Update Oracle Linux 9.0 to latest UEK R7"
+      ansible.builtin.shell: "yum update --nogpgcheck -y"
+      register: kernel_upgrade_result
+      delegate_to: "{{ vm_guest_ip }}"
+
+- name: "Update mainline kernel to {{ upgrade_kernel_version }} for Ubuntu"
+  ansible.builtin.shell: |
+    wget -c -- $(curl -s {{ mainline_kernel_url }}/v{{ upgrade_kernel_version }}/amd64/ | sed 's/"\|>\|</\n/g' | grep deb | uniq | while read line; do echo "{{ mainline_kernel_url }}/v{{ upgrade_kernel_version }}/amd64/${line}";done)
+    if [ $? -eq 0 ]; then dpkg -i linux*.deb; rm -rf linux*.deb; fi
+    sed -i 's/#WaylandEnable=false/WaylandEnable=false/'  /etc/gdm3/custom.conf
```
keirazhang, line 28: This is not a good practice by having so many pipes in one shell command. It is not good at understanding. And it is lack of information for triaging issues. 

By the way for Oracle Linux this task's registration kernel_upgrade_result might overwrite the above one even though it has a when. You'd better to test such scenario with a simple playbook

linux/deploy_vm/upgrade_linux_kernel.yml
```diff
@@ -0,0 +1,72 @@
...
+          - kernel_after_upgrade is version(kernel_before_upgrade, '>')
+        fail_msg: >-
+          Kernel upgrading failed. Before upgrade, the kernel
+          version is '{{ kernel_before_upgrade }}', after upgrade the kernel
+          version is '{{ kernel_after_upgrade }}'.
+
+    - name: "Check gcc version in ubuntu {{ kernel_after_upgrade }}"
+      ansible.builtin.shell: |
+        gcc_ver=$(cat /proc/version | grep -Po 'gcc-\d+')
+        if ! dpkg-query -W -f='${Status}' "$gcc_ver"  | grep "ok installed"; then apt install -y "$gcc_ver"; fi
```
keirazhang, line 69: What if gcc_ver is empty? Please split this task into 2 tasks. The first one is to get gcc version, the second one is to install it.

vars/test.yml
```diff
@@ -415,6 +415,11 @@ flatcar_python3_download_url: "https://downloads.activestate.com/ActivePython/re
 #
 # http_proxy_vm: "http://myproxy.company.com:8080"
 
+# For Linux Mainline Kernel Upgrade only in Ubuntu:
+# e.g., 'latest', '6.10', '6.11-rc1'.
+#
+# upgrade_kernel_version: latest
```
keirazhang, line 421: If mainline kernel versions all have a 'v' prefix, please suggest user to use v6.10, v6.11-rc1, etc

linux/deploy_vm/ubuntu/get_mainline_kernel.yml
```diff
@@ -0,0 +1,45 @@
...
+    return_content: true
+  register: check_kernel_result
+  ignore_errors: yes
+  when: upgrade_kernel_version is match('[0-9]+.[0-9a-z.-]+')
+
+- name: "Get the latest kernel version"
+  when: >
+    (upgrade_kernel_version | lower == 'latest' or
+     upgrade_kernel_version is not match('[0-9]+.[0-9a-z.-]+') or
+     check_kernel_result.status is undefined or check_kernel_result.status != 200)
```
keirazhang, line 20: I wonder why not fail the task when check_kernel_result is failed. Since `upgrade_kernel_version` is from user's input, user is responsible for inputting a valid and correct kernel version. If the kernel version doesn't exists, then fail the test. Otherwise setting `upgrade_kernel_version: latest` is meaningless as user can just input an invalid kernel version it will always get the latest one. Why do we need user to set  `upgrade_kernel_version: latest` then?

linux/deploy_vm/ubuntu/get_mainline_kernel.yml
```diff
@@ -0,0 +1,45 @@
...
+        that:
+          - get_kernel_version_result is defined
+          - get_kernel_version_result.stdout_lines is defined
+          - get_kernel_version_result.stdout_lines | length >= 1
+          - get_kernel_version_result.stdout_lines[0] is match('[0-9]+.[0-9a-z.-]+')
+        fail_msg: "Failed to get the latest version: {{ get_kernel_version_result.stderr }}"
+        success_msg: "Get the latest version: get_kernel_version_result.stdout_lines[0]"
+    - name: "Set the kernel version"
+      ansible.builtin.set_fact:
+        upgrade_kernel_version: "{{ get_kernel_version_result.stdout_lines[0] }}"
```
keirazhang, line 45: Here is my thought about this yaml file:

1. You can add a fact for example `linux_kernel_versions`, which is a list of linux kernel versions. If `upgrade_kernel_version` is a version, then the list only has one item which is the value of `upgrade_kernel_version`; If upgrade_kernel_version is latest, add a following task to get a list of all linux kernels from the mainline kernel url to set `linux_kernel_versions`

For example
```
- name: "Get linux kernel versions"
  ansible.builtin.uri:
    url: "{{ mainline_kernel_url }}"
    return_content: true
  register: get_kernel_versions
```
Then you can use `regex_findall` to get all linux kernel versions from `get_kernel_versions.content` , like
```
linux_kernel_versions: "{{ get_kernel_versions.content | regex_findall('v\\d+[\\w\\.-]*') | reverse }}"
```
You can use `linux_kernel_versions[0:10]` to get latest 10 versions .
```
linux_kernel_versions: "{{ linux_kernel_versions[0:10] }}"
```

Then use the list to loop
``
- name: "Check the kernel version exists"
  ansible.builtin.uri:
    url: "{{ mainline_kernel_url }}/v{{ upgrade_kernel_version }}/amd64"
    return_content: true
  register: check_kernel_result
  ignore_errors: yes
  with_items: "{{ linux_kernel_versions }}"
```
You can check the result of check_kernel_result for each item, and you can get kernel deb files from each item's content by using regex_findall or other filters.

...

Things can be easier.

Comment (keirazhang) [reply linux/deploy_vm/ubuntu/get_mainline_kernel.yml:45]: Further more, below task can be put into a separate yaml file(e.g. get_linux_kernel.yml), used for checking and even downloading linux deb files. 

```
name: "Check the kernel version exists"
ansible.builtin.uri:
url: "{{ mainline_kernel_url }}/v{{ linux_kernel_version }}/amd64"
return_content: true
register: check_kernel_result
ignore_errors: yes
```

In this file, mainly two facts are needed, one for founded linux deb files urls, the other for indicating whether those files were downloaded successfully (presume linux_kernel_downloaded). And it can have a get_url task for downloading all of these files. After that, set linux_kernel_downloaded to true

After `linux_kernel_versions` is set, you can include this separate file looped with `linux_kernel_versions`
```
    - name: "Get linux kernel {{ upgrade_kernel_version }}"
      include_tasks:
        file: get_linux_kernel.yml
        apply:
           when: linux_kernel_downloaded is undefined or not linux_kernel_downloaded
      with_list: "{{ linux_kernel_versions }}"
      loop_control:
         loop_var: linux_kernel_version
```

Then the loop will be break when it found the first valid linux kernel version.

After that, if linux_kernel_downloaded is true, it means there is kernel downloaded, then you can install it in following tasks; otherwise, it means no valid kernel found, it should fail the test.

PR #619, opened 2024-08-06
[Windows] change disk partition size in Autounattend xml files

Tomorrow9:
1. The size of the MSR is 16 MB.
https://learn.microsoft.com/en-us/windows-hardware/manufacture/desktop/configure-uefigpt-based-hard-drive-partitions?view=windows-11
2. Advanced Format 4K Native (4Kn) drives are supported on UEFI-based computers only.
https://learn.microsoft.com/en-us/windows-hardware/manufacture/desktop/hard-drives-and-partitions?view=windows-11
3. For Advanced Format 4K Native drives (4-KB-per-sector) drives, the minimum partition size is 260 MB, due to a limitation of the FAT32 file format.


keirazhang (approved):
LGTM. Thanks!

PR #630, opened 2024-09-19
[Linux] Fix issue caused by DNF version bump on Fedora 41

keirazhang:
In Fedora 41, DNF version is 5.2.5.0, which requires to use Ansible dnf5 module for package management. But currently dnf5 module is still under development, and would throw exception on Fedora 41. So this fix is to replace dnf or dnf5 module with dnf commands as a workaround.


linux/setup/reconfig_existing_guest.yml
```diff
@@ -37,7 +53,7 @@
     - name: "Install 'sudo', 'gawk', 'python3-rpm', 'tar' on VMware Photon OS"
```
123lzxm, line 53: 'python3-rpm' can also be removed from the task name

linux/utils/install_uninstall_package.yml
```diff
@@ -130,34 +134,24 @@
...
               ansible.builtin.set_fact:
-                dnf_major_version: "{{ dnf_major_version_result.stdout }}"
-              when:
-                - dnf_major_version_result.rc is defined
-                - dnf_major_version_result.rc | int == 0
-                - dnf_major_version_result.stdout is defined
-                - dnf_major_version_result.stdout
+                package_manage_cmd: |-
+                  {%- if package_state in [ 'installed', 'present', 'latest' ] -%}
+                    dnf install -y {{ package_manage_list_str }}
```
123lzxm, line 148: Should it use {{ guest_os_ansible_pkg_mgr }} instead of directly setting it to "dnf"?

keirazhang: No, it can't. guest_os_ansible_pkg_mgr could be dnf or dnf5, but the command should be dnf.

linux/setup/reconfig_existing_guest.yml
keirazhang: Fixed. Thanks!

123lzxm (approved):

PR #641, opened 2024-10-29
Set vmtools_is_installed with vm_tools_version_status

keirazhang:
`vm_tools_install_status` was removed from vmtools_info. This fix is to update `vmtools_is_installed` with `vmtools_info.vm_tools_version_status`


Tomorrow9 (approved):
LGTM, thanks!

PR #646, opened 2024-11-01
[FreeBSD]power_operation_scripts failed on FreeBSD 13.x and 14.x with OVT installed from source

ZouYuhua:
Problem:
power_operation_scripts failed on FreeBSD 13.x and 14.x with OVT installed from source

Resolution:
1. FreeBSD VM has OVT packages installed, and then it will uninstall original OVT packages and install with OVT from source. In such scenario, our case will firstly download the original vmtools service file before uninstalling the OVT packages, and then reuse them after OVT source install. Then the original vmtools file should be updated with /etc/vmware-tools/tools.conf before using it. You may update it at https://github.com/vmware/ansible-vsphere-gos-validation/blob/main/linux/open_vm_tools/add_enable_service.yml#L63

2. FreeBSD VM doesn't have OVT packages installed, and then it will install OVT from source. In such scenario, our case will create the vmtools service file from a template https://github.com/vmware/ansible-vsphere-gos-validation/blob/main/linux/open_vm_tools/templates/vmware-guestd.tmpl#L31. In such case, you can directly update this template file with /etc/vmware-tools/tools.conf

Test result:
Test passed with bundled tools & 12.5.0 tools from source build in FreeBSD 13

### DIFF
--- a/linux/open_vm_tools/add_enable_service.yml
+++ b/linux/open_vm_tools/add_enable_service.yml
@@ -71,7 +71,7 @@
           delegate_to: "{{ vm_guest_ip }}"
           ignore_errors: true
 
-        - name: "Update tools.conf path when install tools from source code"
+        - name: "Update tools.conf path to /etc/vmware-tools/tools.conf if exists"
           ansible.builtin.replace:
             path: "{{ service_file }}"
             regexp: '/usr/local/share/vmware-tools/tools.conf'
--- a/linux/open_vm_tools/add_enable_service.yml
+++ b/linux/open_vm_tools/add_enable_service.yml
@@ -56,30 +56,18 @@
         service_enabled: true
         service_state: "started"
 
-- name: "Enable FreeBSD service {{ service_name }}"
+- name: "Update tools.conf path and enable FreeBSD service {{ service_name }}"
   when:
     - guest_os_family == 'FreeBSD'
     - service_name == 'vmware-guestd'
   block:
-    - name: "Update tools.conf path for FreeBSD when tools is installed from source code"
+    - name: "Update tools.conf path to /etc/vmware-tools/tools.conf when tools is installed from source code"
+      ansible.builtin.replace:
+        path: "{{ service_file }}"
+        regexp: '/usr/local/share/vmware-tools/tools.conf'
+        replace: '/etc/vmware-tools/tools.conf'
+      delegate_to: "{{ vm_guest_ip }}"
       when: linux_ovt_install_type == 'source'
-      block:
-        - name: "Check file /etc/vmware-tools/tools.conf exists"
-          ansible.builtin.stat:
-            path: "/etc/vmware-tools/tools.conf"
-          register: tools_config_file_stat_result
-          delegate_to: "{{ vm_guest_ip }}"
-          ignore_errors: true
-
-        - name: "Update tools.conf path to /etc/vmware-tools/tools.conf if exists"
-          ansible.builtin.replace:
-            path: "{{ service_file }}"
-            regexp: '/usr/local/share/vmware-tools/tools.conf'
-            replace: '/etc/vmware-tools/tools.conf'
-          delegate_to: "{{ vm_guest_ip }}"
-          when:
-            - tools_config_file_stat_result.stat.exists is defined
-            - tools_config_file_stat_result.stat.exists
 
     - name: "Enable FreeBSD service {{ service_name }}"
       include_tasks: ../utils/replace_or_add_line_in_file.yml
--- a/linux/open_vm_tools/add_enable_service.yml
+++ b/linux/open_vm_tools/add_enable_service.yml
@@ -61,6 +61,26 @@
     - guest_os_family == 'FreeBSD'
     - service_name == 'vmware-guestd'
   block:
+    - name: "Update tools.conf file for FreeBSD when tools is installed from source code"
+      when: linux_ovt_install_type == 'source'
+      block:
+        - name: "Check file /etc/vmware-tools/tools.conf exists"
+          ansible.builtin.stat:
+            path: "/etc/vmware-tools/tools.conf"
+          register: tools_config_file_stat_result
+          delegate_to: "{{ vm_guest_ip }}"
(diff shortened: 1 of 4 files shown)


ZouYuhua:
The testing on FreeBSD 14 is on-going ...

ZouYuhua:
> The testing on FreeBSD 14 is on-going ...

Test passed with bundled tools & 12.5.0 tools from source build in FreeBSD 13 & FreeBSD 14

linux/open_vm_tools/add_enable_service.yml
```diff
@@ -56,11 +56,19 @@
...
-- name: "Enable FreeBSD service {{ service_name }}"
+- name: "Update tools.conf path and enable FreeBSD service {{ service_name }}"
   when:
     - guest_os_family == 'FreeBSD'
     - service_name == 'vmware-guestd'
   block:
+    - name: "Update tools.conf path to /etc/vmware-tools/tools.conf when tools is installed from source code"
+      ansible.builtin.replace:
+        path: "{{ service_file }}"
+        regexp: '/usr/local/share/vmware-tools/tools.conf'
```
keirazhang, line 67: Do not hard code the tools.conf path. Please use a regex pattern like: `-c .*tools.conf`

linux/open_vm_tools/add_enable_service.yml
```diff
@@ -56,11 +56,19 @@
...
+        replace: '/etc/vmware-tools/tools.conf'
+      delegate_to: "{{ vm_guest_ip }}"
+      when: linux_ovt_install_type == 'source'
```
keirazhang, line 70: Only OVT source install will execute tasks in this file. So here doesn't need to check `linux_ovt_install_type`. But you need to check `service_file_local_path` because only when the service file is copied from an original one, the tools.conf path need to be updated.
```
    - service_file_local_path is defined
    - service_file_local_path
 ```

For service file created from the template, it has correct tools.conf path as you updated

File: linux/power_operation_scripts/set_power_script.yml [hunk 3]
```diff
@@ -115,26 +115,54 @@
...
+            fail_msg: "Failed to find the {{ script_file_path }} in {{ vmtools_vmsvc_log_file }} for {{ power_script_type }} {{ power_script_op }}."
+            success_msg: "Successfully found the {{ script_file_path }} in {{ vmtools_vmsvc_log_file }} for {{ power_script_type }} {{ power_script_op }}."
+   
+    - name: "Check the log file in {{ guest_os_ansible_distribution }}"
+      when: guest_os_ansible_distribution == 'FreeBSD'
+      block:
+        - name: "Search the log file {{ script_file_path }} in {{ guest_os_ansible_distribution }}"
+          ansible.builtin.shell: "grep {{ script_file_path }} *vmsvc*.log ./vmware-tools/*.log"
+          args:
+            chdir: "/var/log"
```
keirazhang, line 148: Why does FreeBSD have a different log directory when tools.conf has below content:
```
[logging]
log=true
vmtoolsd.level=debug
vmtoolsd.handler=file
vmtoolsd.data=/var/log/vmware-tools/vmtoolsd.log
vmsvc.level=debug
vmsvc.handler=file
vmsvc.data=/var/log/vmware-tools/vmsvc.log
```

linux/utils/enable_vmtools_logging.yml
```diff
@@ -22,10 +22,12 @@
 # VMware Tools config file on FreeBSD with open-vm-tools 10.1.x or later
 - name: "Set fact of VMware Tools config file for FreeBSD"
   ansible.builtin.set_fact:
-    vmtools_config_file: "/usr/local/share/vmware-tools/tools.conf"
+    vmtools_config_file: |-
+      {%- if linux_ovt_install_type is defined and linux_ovt_install_type == 'source' -%}/etc/vmware-tools/tools.conf
```
keirazhang, line 26: Please use `vmtools_install_type` instead of `linux_ovt_install_type` as `vmtools_install_type` will always be set but `linux_ovt_install_type` not.

You can just change the when at line 19~20 and here. VMTools 10.1.0 is deprecated. You can replace the condition about checking vmtools_version with checking `vmtools_install_type`.

For example, line 20 can be replaced by: `vmtools_install_type != 'package'`. And add `vmtools_install_type == 'package'` for this task.

ZouYuhua: Is it OK for two tasks: one for linux and other for FreeBSD ?

linux/utils/enable_vmtools_logging.yml
```diff
@@ -11,21 +11,20 @@
...
-# VMware Tools config file on FreeBSD with open-vm-tools 10.1.x or later
+# VMware Tools config file on FreeBSD
 - name: "Set fact of VMware Tools config file for FreeBSD"
   ansible.builtin.set_fact:
-    vmtools_config_file: "/usr/local/share/vmware-tools/tools.conf"
-  when:
-    - guest_os_ansible_distribution == 'FreeBSD'
-    - vmtools_version is version('10.1.0', '>=')
+    vmtools_config_file: |-
+      {%- if vmtools_install_type is defined and vmtools_install_type == 'source' -%}/etc/vmware-tools/tools.conf
```
keirazhang, line 24: Here doesn't need to check `vmtools_install_type is defined` as it is defined with `vmtools_version` in the same task file. And there could be 3 possible values for it: unknown, source, and package. Only when `vmtools_install_type == 'package'` we can assure its tools conf is `/usr/local/share/vmware-tools/tools.conf`. For the other two values, we can assume it is `/etc/vmware-tools/tools.conf`.

linux/power_operation_scripts/set_power_script.yml
ZouYuhua: Change to only check ../vmware-tools/.log. Testing is on-going...

ZouYuhua:
@keirazhang 
Pass the testing with bundled & 12.5.0 tools in FreeBSD 13 and FreeBSD 14 after code review

keirazhang (approved):
LGTM. Thanks!

PR #651, opened 2024-11-19
Add owner and group for creating local directory

keirazhang:
Added two options for creating local directory: dir_owner and dir_group. So that we can change the owner and group of the new created directory

### DIFF
--- a/common/create_directory.yml
+++ b/common/create_directory.yml
@@ -5,7 +5,7 @@
 # Parameters:
 #   dir_path: the new directory path to be created.
 #   dir_mode (Optional): the permission of new directory, default is '0755'.
-#   dir_owner (Optional): the user name that owne the new directory.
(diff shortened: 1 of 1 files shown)


common/create_directory.yml
```diff
@@ -3,13 +3,17 @@
 ---
 # Create new directory with specified path
 # Parameters:
-#   dir_path: the new directory path to be created
-#   dir_mode: the permission of new directory, default is '0755'
+#   dir_path: the new directory path to be created.
+#   dir_mode (Optional): the permission of new directory, default is '0755'.
+#   dir_owner (Optional): the user name that owne the new directory.
```
Tomorrow9, line 8: owne -> own

Tomorrow9 (approved):

common/create_directory.yml
keirazhang: Thanks!

PR #659, opened 2024-12-05
Update deploy_vm test case full name in log plugin

keirazhang:
If test failed before setting deploy_vm test case name, changes in this fix will update the test case name as possible as the plugin can.


plugin/ansible_vsphere_gosv_log.py
```diff
@@ -780,6 +780,21 @@ def _print_test_results(self):
         test_idx = 0
         for test_id in self.test_runs:
             test_result = self.test_runs[test_id]
+            if test_result.name == 'deploy_vm':
```
Tomorrow9, line 783: And `new_vm` is set to true.

Tomorrow9 (approved):
LGTM, thanks!

PR #662, opened 2024-12-24
Increase kern.kstack_pages for 32bit FreeBSD

keirazhang:
This is to work around core dump issue of 32bit FreeBSD VM on Intel Sapphire Rapids CPU based on FreeBSD's solution:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=283276#c1

### DIFF
--- a/linux/deploy_vm/templates/post_install_scripts/freebsd_post_config.sh
+++ b/linux/deploy_vm/templates/post_install_scripts/freebsd_post_config.sh
@@ -133,6 +133,10 @@ sed -i '' 's/kern.geom.label.gptid.enable="0"/kern.geom.label.gptid.enable="1"/'
 echo 'kern.geom.label.gpt.enable="1"' >>/boot/loader.conf
 echo 'kern.geom.label.ufs.enable="1"' >>/boot/loader.conf
 echo 'kern.geom.label.ufsid.enable="1"' >>/boot/loader.conf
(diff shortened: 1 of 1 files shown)


linux/deploy_vm/templates/post_install_scripts/freebsd_post_config.sh
```diff
@@ -133,6 +133,10 @@ sed -i '' 's/kern.geom.label.gptid.enable="0"/kern.geom.label.gptid.enable="1"/'
 echo 'kern.geom.label.gpt.enable="1"' >>/boot/loader.conf
 echo 'kern.geom.label.ufs.enable="1"' >>/boot/loader.conf
 echo 'kern.geom.label.ufsid.enable="1"' >>/boot/loader.conf
+if [ "$machtype" != "amd64" ] && [ "$machtype" != "x86_64" ]; then
```
123lzxm, line 136: Is Intel Sapphire Rapids CPU also x86_64?

123lzxm: Except amd64 and x86_64, what other types are there?

keirazhang: `amd64` and `x86_64` is to identify 64bit guest OS not the CPU model.

123lzxm (approved):

PR #672, opened 2025-02-07
Set ESXi CPU code name from EVC mode key

keirazhang:
The search result page of Intel CPU names is dynamic loaded now, and can't be parsed any more. By checking ESXi properties, I found summary.maxEVCModeKey can reflect the CPU code name. So I updated the CPU code name by using `summary.maxEVCModeKey`. The limitation of this key is that it only exists when VC is defined. Since our testing is mainly run with VC+ESXi, so it should be acceptable.


Tomorrow9 (approved):
LGTM, thanks!

PR #674, opened 2025-02-17
[FreeBSD 13] Apply dndcp patch for compiling open-vm-tools

keirazhang:
Test case ovt_verify_src_install failed on FreeBSD 13.5 with below compiling error:

```
In file included from dndGuestBase/copyPasteDnDWrapper.cpp:30:
In file included from dndGuestBase/copyPasteDnDX11.h:30:
In file included from dndGuestBase/dndUIX11.h:30:
In file included from ./stringxx/string.hh:34:
/usr/include/c++/v1/string:820:42: error: implicit instantiation of undefined template 'std::char_traits<unsigned short>'
  820 |   static_assert(is_same<_CharT, typename traits_type::char_type>::value,
      |                                          ^
./stringxx/string.hh:270:18: note: in instantiation of template class 'std::basic_string<unsigned short>' requested here
  270 |    return string(lhs) + rhs;
      |                  ^
/usr/include/c++/v1/__fwd/string.h:23:29: note: template is declared here
   23 | struct _LIBCPP_TEMPLATE_VIS char_traits;
      |                             ^
1 error generated.
```
After applying FreeBSD's patch /usr/ports/emulators/open-vm-tools/files/patch-services_plugins_dndcp_stringxx_string.hh, this issue disappeared.

So we will apply this patch before compiling open-vm-tools on FreeBSD 13


123lzxm (approved):
LGTM, thanks.

PR #677, opened 2025-02-24
[Windows] add condition of NVME known issue on Windows Server 2025

Tomorrow9:
This NVMe known issue may also happen when adding a new NVMe disk to the boot NVMe disk controller with EFI firmware sometime, so add this condition.


keirazhang (approved):
LGTM. Thanks!

PR #678, opened 2025-02-25
[Windows] add VMCI known issue on ESXi 8.0.x

Tomorrow9:


windows/setup/handle_vmci_known_issue.yml
```diff
@@ -0,0 +1,55 @@
...
+        name: "{{ vm_name }}"
+      register: retry_vmtools_status
+      ignore_errors: true
+      until:
+        - retry_vmtools_status is defined
+        - retry_vmtools_status.vmtools_info is defined
+        - retry_vmtools_status.vmtools_info.vm_tools_running_status is defined
+        - retry_vmtools_status.vmtools_info.vm_tools_running_status == 'guestToolsNotRunning'
+      retries: 60
+      delay: 5
```
keirazhang, line 35: I think this task can be replaced by `common/vm_wait_vmtools_status.yml`, but the `vmtools_is_running` need to be saved in a temporary variable to use as the when.

windows/setup/handle_vmci_known_issue.yml
```diff
@@ -0,0 +1,55 @@
...
+    - name: "Meet known issue conditions"
+      ansible.builtin.set_fact:
+        vmci_known_issue: "{{ not vmtools_is_running or (retry_vmtools_status is defined and retry_vmtools_status.vmtools_info.vm_tools_running_status | default('') == 'guestToolsNotRunning') }}"
+
+    - name: "Known issue - Guest OS hang after reverting to snapshot when VM VBS enabled"
+      ansible.builtin.debug:
+        msg:
+          - "VMware Tools in guest OS turns to be not running and guest OS hang after reverting to snapshot when VM VBS is enabled."
+          - "This is a known issue on ESXi 8.0U1 to 8.0U3c and VM hardware version > 20."
+          - "Fixes are in the following ESXi release, e.g., new patch release of ESXi 8.0.3."
```
keirazhang, line 47: Is there a KB link can be added in this message?

windows/setup/handle_vmci_known_issue.yml
```diff
@@ -0,0 +1,55 @@
...
+      tags:
+        - known_issue
+      when: vmci_known_issue
+
+    - name: "Exit testing due to guest OS hang caused by known issue"
+      ansible.builtin.fail:
+        msg: "Testing failed because test environment meets the VMCI known issue conditions on ESXi {{ esxi_version }} build {{ esxi_build }}."
```
keirazhang, line 54: This fail message can be more specific, like:
```
Testing failed because guest OS hang after reverting to snapshot when VM VBS is enabled, which was caused by a VMCI known issue on ESXi {{ esxi_version }} build {{ esxi_build }}.
```
This will be more meaningful for auto triage.

windows/setup/test_setup.yml
```diff
@@ -18,6 +18,15 @@
...
   when: vm_settings_checked is undefined
 
+- name: "Handle VMCI known issue on ESXi 8.0.x"
+  include_tasks: handle_vmci_known_issue.yml
+  when:
+    - guest_os_ansible_architecture == "64-bit"
+    - vm_hardware_version_num | int > 20
+    - esxi_version is version('8.0.0', '>')
+    - esxi_version is version('8.0.3', '<=')
+    - esxi_build | int <= 24414501
```
keirazhang, line 28: I think here can also add checking ```base_snapshot_exists```. If it is false, `base_snapshot_check_revert.yml` won't revert to base snapshot. In such case, VM won't hit the known issue.

windows/setup/handle_vmci_known_issue.yml
Tomorrow9: Yeah, right, thanks.

windows/setup/handle_vmci_known_issue.yml
Tomorrow9: No KB link.

windows/setup/test_setup.yml
Tomorrow9: Added, thanks.

windows/setup/handle_vmci_known_issue.yml
```diff
@@ -0,0 +1,61 @@
...
+        - name: "Save VMware Tools running status"
+          ansible.builtin.set_fact:
+            vmtools_is_running_after: false
+          when:
+            - get_vmtools_info.vmtools_info.vm_tools_running_status is defined
+            - get_vmtools_info.vmtools_info.vm_tools_running_status == 'guestToolsNotRunning'
+
+    - name: "Meet known issue conditions"
+      ansible.builtin.set_fact:
+        vmci_known_issue: "{{ (vmtools_is_installed_before and not vmtools_is_running_before) or (not vmtools_is_running_after | default(true)) }}"
```
keirazhang, line 43: `vmtools_is_running_after | default(true)` The default value is kind of misleading literally. Afterall, tools could be not installed.  I suggest setting vmci_known_issue in tasks of line 18 and line 36 separately: 
In line 18 task:
```
vmci_known_issue: "{{ vmtools_is_installed_before and not vmtools_is_running_before }}"
```
In line 36 task:
```
vmci_known_issue: true
```

common/test_rescue.yml
```diff
@@ -76,21 +76,27 @@
...
-        - datastore_file_download_success
+    - name: "Handle Windows guest OS situation"
+      when: gosv_test_suite == 'windows'
       block:
-        - name: "Look up Windows guest BSOD in vmware.log"
+        - name: "Set exit testing when meets VMCI known issue"
           ansible.builtin.set_fact:
-            winbsod_in_vmware_log: "{{ lookup('file', datastore_file_download_result.dest) | regex_findall('.*WinBSOD:.*') }}"
+            exit_testing_when_fail: true
+          when: vmci_known_issue | default(false)
```
keirazhang, line 85: It's better to put this task into the test case not this common task. Otherwise, all failed test cases will check the value of vmci_known_issue.

Tomorrow9: Since this failure would happen in the test setup in any test case, and if this known issue happens, there is no need to run the following test cases.

keirazhang: I mean set ```exit_testing_when_fail: true``` in ```handle_vmci_known_issue.yml```  when vmci_known_issue is set to true . After all, the known issue does not happen on all ESXi  servers.

Tomorrow9:
Hold on this change since this issue should not be reproduced internally or externally. Thanks.

PR #683, opened 2025-03-13
Replace "include" with "include_tasks" in testcase power_operation_scripts

ZouYuhua:
ansible.builtin.include has been removed and use include_tasks instead for testcase power_operation_scripts

Test passed:
image


keirazhang (approved):
LGTM. Thanks!

PR #684, opened 2025-03-14
[Linux] Update GOSC network config file for SLES 16

keirazhang:
SLES 16 is using NetworkManager to manage network connections. This change is to get the connection file by nmcli command on SLES 16.


Tomorrow9 (approved):
LGTM, thanks!

PR #691, opened 2025-03-27
[Windows] Change to fetching GOSC log files using vm_guest_file_operation.yml in DHCP GOSC

Tomorrow9:
Since there is wrong used loop var when there are nested loops, here to change to use the common task `vm_guest_file_operation.yml ` to fetch the GOSC log files no matter it's static IP GOSC or DHCP GOSC.


windows/utils/win_get_file_folder.yml
```diff
@@ -3,29 +3,43 @@
...
+    - name: "Get specified file from guest OS"
+      include_tasks: win_get_file.yml
+      vars:
+        win_get_file_src_path: "{{ win_src_path }}"
+        win_get_file_dst_path: "{{ win_dst_path }}"
+        win_get_file_ignore_error: true
+      when: not win_is_folder_result
+
+    - name: "Get files in specified folder from guest OS recursively"
+      include_tasks: win_get_folder.yml
```
keirazhang, line 41: I saw `win_get_folder.yml` included `win_get_file_folder.yml` and `win_get_file_folder.yml` included `win_get_folder.yml`.  Won't the insider vars changing affect outsider tasks?

I think it's better to get a list of files' full paths under this folder and all its sub folders, then use `win_get_file.yml` to fetch them one by one.

Tomorrow9: I removed these 2 utils files since they're not used in test cases. If there is new request then will add new tasks with `ansible.windows.win_find` module it supports 'get files recursively' now. Thanks.

windows/utils/win_is_folder.yml
```diff
@@ -3,26 +3,28 @@
...
+- name: "Initialize the check result"
   ansible.builtin.set_fact:
     win_is_folder_result: false
 
-# Check if specified path is a directory in Windows guest OS
-- include_tasks: win_execute_cmd.yml
+- name: "Check if specified path is a directory in Windows guest OS"
+  include_tasks: win_execute_cmd.yml
   vars:
     win_powershell_cmd: "(Get-Item -Path '{{ win_is_folder_path }}') -is [System.IO.DirectoryInfo]"
```
keirazhang, line 18: Does here not need `win_execute_cmd_ignore_error: true` any more?

Tomorrow9: Add it back in case removing will affect the tasks which includes this utility. Thanks.

keirazhang (approved):
LGTM. Thanks!

PR #704, opened 2025-05-07
[Linux] Get installed open-vm-tools packages before uninstalling

keirazhang:
openSUSE 16.0 Beta release doesn't have open-vm-tools-desktop installed as it has no desktop. To avoid uninstalling non-existing package, this fix filtered out installed open-vm-tools packages, and then uninstall them.

```
+---------------------------------------------------------------------+
| GuestInfo Detailed Data   | architecture='X86'                      |
|                           | bitness='64'                            |
|                           | cpeString='cpe:/o:opensuse:leap:16.0'   |
|                           | distroAddlVersion='16.0 Beta'           |
|                           | distroName='openSUSE Leap'              |
|                           | distroVersion='16.0'                    |
|                           | familyName='Linux'                      |
|                           | kernelVersion='6.12.0-160000.9-default' |
|                           | prettyName='openSUSE Leap 16.0 Beta'    |
+---------------------------------------------------------------------+

Test Results (Total: 1, Passed: 1, Elapsed Time: 00:08:56)
+--------------------------------------------------+
| ID | Name                   | Status | Exec Time |
+--------------------------------------------------+
|  1 | ovt_verify_pkg_install | Passed | 00:08:35  |
+--------------------------------------------------+
```

### DIFF
--- a/linux/open_vm_tools/uninstall_ovt.yml
+++ b/linux/open_vm_tools/uninstall_ovt.yml
@@ -5,26 +5,26 @@
 # Parameters:
 #   ovt_packages: a list of open-vm-tools packages
 #
+
+- name: "Get installed packages on {{ vm_guest_os_distribution }}"
+  include_tasks: ../utils/get_installed_packages.yml
+
+- name: "Set fact of installed open-vm-tools related packages"
+  ansible.builtin.set_fact:
+    installed_ovt_packages: "{{ ovt_packages | select('in', guest_installed_packages) }}"
+
+- name: "Display open-vm-tools related packages which are installed in guest OS"
+  ansible.builtin.debug: var=installed_ovt_packages
+
 # Uninstall open-vm-tools packages
-- name: "Uninstall packages {{ ovt_packages }}"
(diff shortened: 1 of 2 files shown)


Tomorrow9:
`ovt_packages` list contains `open-vm-tools-desktop` when `guest_os_with_gui`, so here openSUSE 16.0 Beta release without desktop but getting parameter `guest_os_with_gui` to be set to True?

Tomorrow9:
> We check whether `display-manager.service` is running to decide guest OS has GUI or not. At present, openSUSE 16 has this service running but its desktop is not ready yet. So it added `open-vm-tools-desktop` into OVT package list.

Then the problem is setting parameter `guest_os_with_gui` according to the real situation.

keirazhang:
> Then the problem is setting parameter `guest_os_with_gui` according to the real situation.

The desktop issue would cause `ovt_verify_status` failure, and I will create another pull requests to update desktop detection by checking login session.

Uninstalled packages should not block the testing. We only need to uninstall whatever it is installed in the guest OS.

Tomorrow9 (approved):
LGTM, thanks!

PR #706, opened 2025-05-09
[Linux] Update desktop environment by checking login session

keirazhang:
The display-manager.service is a systemd service responsible for managing the graphical login screen, also known as a display manager (e.g., GDM, LightDM, SDDM). A display manager typically starts the X server (or Wayland) and launches a desktop session, such as GNOME, KDE Plasma, XFCE, etc.

It's possible but rare for the display manager to be running without a full desktop environment (like openSUSE 16.0 Beta). In such case, checking display-manager.service is not enough. This fix added login session checking. If there is x11 or wayland login session in guest OS, we will consider it has a desktop environment.


keirazhang:
At present, these changes doesn't help identify desktop issue on Linux. I will reopen it when there is further improvement.

PR #710, opened 2025-05-13
Add getting guest options and setting defualt configurations for new VM in env_setup

Tomorrow9:
1. Add get ESXi supported VM hardware versions list in env setup stage.
2. Add get guest options before deploying new VM to check whether configured boot disk controller is supported.
3. Add get guest options once in test setup for guest supported info will be used in the following test cases.


env_setup/env_setup.yml
```diff
@@ -65,31 +68,8 @@
...
-
-        - name: "Power on VM"
-          include_tasks: ../common/vm_set_power_state.yml
-          vars:
-            vm_power_state_set: "powered-on"
-          when: vm_power_state_get == "poweredOff"
-
-        - name: "Wait for VM primary network adapter has MAC address"
-          include_tasks: ../common/vm_wait_primary_nic_mac.yml
+      include_tasks: handle_existing_vm.yml
```
keirazhang, line 71: Using `handle` looks like there is some problem with the VM. How about changing the file name to `setup_existing_vm.yml` or `existing_vm_setup.yml`?

linux/deploy_vm/deploy_vm_from_iso.yml
```diff
@@ -3,19 +3,57 @@
...
+      {{
+         (network_adapter_type is defined and network_adapter_type) 
+         | ternary(network_adapter_type | lower, guest_config_options.default_ethernet) 
+      }}
+    firmware: >-
+      {{
+         (firmware is defined and firmware) 
+         | ternary(firmware | lower, guest_config_options.rec_firmware) 
+      }} 
     secureboot_enabled: "{{ secureboot_enabled is defined and secureboot_enabled | bool }}"
```
keirazhang, line 34: If `secureboot_enabled is not set`, shall here use `guest_config_options.default_secure_boot`?

windows/deploy_vm/deploy_vm_from_iso.yml
```diff
@@ -4,19 +4,53 @@
...
+      New VM boot disk controller '{{ boot_disk_controller }}' is not in the disk controller
+      support list in guest config options '{{ guest_config_options.support_disk_controller }}'.
+
+- name: "Check new VM network adatper is supported"
+  ansible.builtin.assert:
+    that:
+      - network_adapter_type in guest_config_options.support_ethernet_card
+    fail_msg: >-
+      New VM network adatper '{{ network_adapter_type }}' is not in the Ethernet support list in guest
+      config options '{{ guest_config_options.support_ethernet_card }}'.
```
keirazhang, line 52: How about moving these tasks for setting VM's hardware version, boot disk controller, ethernet, firmware, etc into a common task, which can be directly included in linux and windows? Or set these default values in env_setup for new_vm = true?

env_setup/handle_existing_vm.yml
```diff
@@ -0,0 +1,26 @@
+# Copyright 2025 VMware, Inc.
+# SPDX-License-Identifier: BSD-2-Clause
+---
+- name: "Not cleanup old snapshot by default"
```
keirazhang, line 4: I wonder whether we can revert the base snapshot firstly if it exists for existing VM. Then no matter old snapshots are cleaned or not, below `vm_get_vm_info.yml` will always be the VM info at base snapshot or the VM info without base snapshot. And we don't need to get it again in `test_setup.yml`, and we can also move `esxi_get_guest_config_options.yml` from `test_setup.yml` to this file.

env_setup/env_setup.yml
Tomorrow9: Changed, thanks.

env_setup/handle_existing_vm.yml
Tomorrow9: This we can take a look at all the related tasks and consider later. Thanks.

env_setup/set_default_vm_vars.yml
```diff
@@ -149,17 +140,14 @@
...
   when: cpu_number is undefined or not cpu_number
 
-- name: "Check VM CPU socket number"
+- name: "Check Windows VM CPU socket number"
   include_tasks: check_cpu_socket.yml
-
-- name: "Set fact of the VM datastore path"
-  ansible.builtin.set_fact:
-    vm_datastore_path: "/vmfs/volumes/{{ datastore }}"
+  when: testing_testcase_file | dirname | basename == "windows"
```
keirazhang, line 145: We have a variable `gosv_test_suite` defined in `main.yml`, here can change to `gosv_test_suite == "windows"`

keirazhang (dismissed):
I have no more comments. Thanks!

123lzxm (dismissed):

keirazhang (dismissed):
LGTM. Thanks!

keirazhang (approved):
LGTM. Thanks!

PR #713, opened 2025-05-14
[Linux] Update VMware Tools version and build number

keirazhang:
VMware Tools  version and build format is changed. This fix is to align this change.


Tomorrow9 (approved):
LGTM, thanks!

PR #720, opened 2025-06-03
[Linux] Support open-vm-tools source install on RHEL familly and VMware Photon OS

keirazhang:
This fix is to support installing open-vm-tools from source on RHEL family like RHEL, AlmaLinux, Rocky Linux, Oracle Linux, MIRACLE LINUX, etc, and to support installing open-vm-tools from source on VMware Photon OS.


Tomorrow9 (approved):
LGTM, thanks!

PR #746, opened 2025-06-30
[Linux] Add LSILogicSAS known issue and workaround for SLES 16

keirazhang:
Test case lsilogicsas_vhba_device_ops failed occasionally because the guest OS cannot recognize the new disk hot added to an existing lsilogicsas controller. It is a Linux kernel issue like Ubuntu. This fix is to mark it as known issue on SLES 16.0 or later and reboot the guest OS to recognize the disk as a workaround.


Tomorrow9 (approved):
LGTM, thanks!

PR #761, opened 2025-07-25
[Linux] Handle SATA hot remove known issue for RHEL family OS

keirazhang:
On RHEL 9.6 and 10.0, there is a config `alpmalpm=medium_power` in its tuned config file, which leads to guest OS can still lists a SATA disk which has been hot removed.  Some other RHEL family OS like CentOS Stream, Oracle Linux, Rocky, AlmaLinux, MIRACLE LINUX, etc also have such issue. 

This fix added the tuned config check and handled this SATA hot remove issue on RHEL family OS.

```
| GuestInfo Detailed Data   | architecture='Arm'                                     |
|                           | bitness='64'                                           |
|                           | cpeString='cpe:/o:redhat:enterprise_linux:10::baseos'  |
|                           | distroAddlVersion='10.0 (Coughlan)'                    |
|                           | distroName='Red Hat Enterprise Linux'                  |
|                           | distroVersion='10.0'                                   |
|                           | familyName='Linux'                                     |
|                           | kernelVersion='6.12.0-55.9.1.el10_0.aarch64'           |
|                           | prettyName='Red Hat Enterprise Linux 10.0 (Coughlan)'  |
+------------------------------------------------------------------------------------+

Test Results (Total: 1, Passed: 1, Elapsed Time: 00:30:14)
+------------------------------------------------+
| ID | Name                 | Status | Exec Time |
+------------------------------------------------+
|  1 | sata_vhba_device_ops | Passed | 00:25:10  |
+------------------------------------------------+
```

### DIFF
--- a/linux/vhba_hot_add_remove/handle_sata_known_issue.yml
+++ b/linux/vhba_hot_add_remove/handle_sata_known_issue.yml
@@ -3,14 +3,47 @@
 ---
 # Handle SATA known issues
 #
-# Pardus 21 XFCE can't detect SATA disk change after hot-remove
-# Work around it by updating the disk state
-- name: "Handle SATA known issue on {{ vm_guest_os_distribution }}"
+# Pardus 21 XFCE can't detect SATA disk change after hot remove
+- name: "Set fact of guest OS has SATA known issue on {{ vm_guest_os_distribution }}"
+  ansible.builtin.set_fact:
+    sata_known_issue_exists: true
   when:
     - guest_os_ansible_distribution == 'Pardus GNU/Linux'
     - guest_os_ansible_distribution_major_ver | int == 21
     - guest_os_edition == 'XFCE'
+
+# RHEL family OS of version 9.6, 10.0 or later, CentOS Stream 9 and 10 can't detect SATA disk change
+# after hot remove when 'alpmalpm=medium_power' exists in tuned config
+- name: "Check whether SATA known issue exists on {{ vm_guest_os_distribution }}"
+  when:
+    - guest_os_family == 'RedHat'
+    - guest_os_ansible_distribution_major_ver | int >= 9
+  block:
+    - name: "Check 'alpm=medium_power' in tuned config"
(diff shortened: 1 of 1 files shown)


linux/vhba_hot_add_remove/handle_sata_known_issue.yml
```diff
@@ -3,14 +3,47 @@
...
+            (guest_os_ansible_distribution_major_ver | int == 9) |
+            ternary('/usr/lib/tuned/balanced/tuned.conf',
+                    '/usr/lib/tuned/profiles/balanced/tuned.conf')
+          }}
+
+    - name: "Set fact of guest OS has SATA known issue on {{ vm_guest_os_distribution }}"
+      ansible.builtin.set_fact:
+        sata_known_issue_exists: true
+      when:
+        - check_alpm_result.stdout is defined
```
Tomorrow9, line 39: Does this situation exist that above task failed and 'check_alpm_result.stdout' is not empty?

keirazhang: I can only think of `unreachable`, which will lead to test exit at above task.

Tomorrow9: Then here `ignore_errors: true`can be removed and let the task fail when it fails, or add a `when` this task failed situation with a specific error message.

keirazhang: No, it can not be removed. If the tuned conf file doesn't have this setting, the return code of grep is not 0, which would lead to the task fail, and that is not expected. The known issue exists only when the tuned conf file has this setting, it didn't exist in old versions or may be fixed in future versions. This task is just to check whether the tuned conf has the setting, and it shouldn't affect the test if it doesn't have the setting.

Tomorrow9: Ok, if this does not affect the testing result.

Tomorrow9 (approved):
LGTM, thanks.

PR #768, opened 2025-08-19
[Linux] Fix Debian 13 autoinstall stuck at "Running preseed"

ZouYuhua:
Debian 13 autoinstall stuck at "Running preseed"

**Two issues:**
1) During autoinstall, it should bind the installer environment’s /cdrom to the target system’s /media/cdrom, but the installer did not do so

   (if the ISO file is not mounted, the autoinstall will stuck)

2) CDROM APT source doesn't work even if manually mount an ISO file or use the mount command in a script to mount the ISO file.

**Resolution:**
   bind the installer environment’s /cdrom to the target system’s /media/cdrom
   Then, the autoinstall won't stuck.

   Note:
   For the issue 2, even if the CDROM APT source doesn't work, it won't 
affect the autoinstall. The packages will be installed from online repo.
   

**Test result:** 
   Passed. (check GOSV-5206)

### DIFF
--- a/linux/deploy_vm/templates/post_install_scripts/apt_install_pkgs.sh
+++ b/linux/deploy_vm/templates/post_install_scripts/apt_install_pkgs.sh
@@ -1,24 +1,24 @@
 #Install open-vm-toools from CDROM if it exists
 cdrom_missing_pkgs=""
 for pkg in $required_pkgs; do
-    echo "Searching package $pkg in CDROM" >/dev/ttyS0
+    echo "Searching package $pkg in CDROM"
     apt list $pkg 2>/dev/null | cut -d '/' -f 1 | grep $pkg
     pkg_in_cdrom=$?
     if [ $pkg_in_cdrom -eq 0 ]; then
         echo "Installing pacakge $pkg from CDROM"
         apt install -y $pkg 2>&1
         if [ $? -ne 0 ]; then
-            echo "ERROR: Failed to install package $pkg from CDROM" >/dev/ttyS0
+            echo "ERROR: Failed to install package $pkg from CDROM"
             cdrom_missing_pkgs="$cdrom_missing_pkgs $pkg"
         fi
     else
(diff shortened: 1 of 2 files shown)


linux/deploy_vm/templates/post_install_scripts/apt_install_pkgs.sh
```diff
@@ -31,6 +31,11 @@ if [ "X$cdrom_missing_pkgs" != "X" ]; then
     echo "Updating list of available packages"
     apt update -y 2>&1
 
+    # It gets stuck at searching / installing packages from online repo during autoinstall for Debian 13
+    # Skipping the installation of these packages during autoinstall will not affect the subsequent test cases.
+    if [[ "X$OS_NAME" == "XDebian GNU/Linux" ]] && [[ "X$VERSION_ID" == "X13" ]]; then
+        cdrom_missing_pkgs=""
+    fi
```
keirazhang, line 38: This is not a reasonable fix. Please figure out why CDROM APT source has error "The repository 'cdrom://[Debian GNU/Linux 13.0.0 _Trixie_ - Official amd64 DVD Binary-1 with firmware 20250809-11:21] trixie Release' does not have a Release file" as below:

```
[1;31mError: [0m[1mThe repository 'cdrom://[Debian GNU/Linux 13.0.0 _Trixie_ - Official amd64 DVD Binary-1 with firmware 20250809-11:21] trixie Release' does not have a Release file.[0m
[1mNotice: [0mUpdating from such a repository can't be done securely, and is therefore disabled by default.[0m
[1mNotice: [0mSee apt-secure(8) manpage for repository creation and user configuration details.[0m
```

You can manually check from a Debian 13 VM:
```
apt-cdrom add
apt-get update
```

You can switch to console at autoinstall by pressing CTRL+ALT+F4, and monitor the console output to see what error is reported when it is stuck.

If the CDROM APT source doesn't work, I think you can comment out the CDROM source before adding online repo.

PR #781, opened 2025-09-25
Not ignore errors when getting vCenter version and build info

Tomorrow9:


common/vcenter_get_version_build.yml
123lzxm, line 1: How about preserving the common uitls as it it and adding the check of vCenter build and version info at the beginning of the test case of gosc?

Tomorrow9: Then everywhere when using this info need to check it again. I think it makes no difference letting it fail at the beginning or at where it used. The principle is let failure happens when there is error as early as possible.

keirazhang (approved):
LGTM. Thanks!

PR #814, opened 2025-12-11
[Windows] Add app quiesce snapshot support for Windows Server 2025 with NVMe disk controller

Tomorrow9:


Tomorrow9:
Defer this change due to another NVMe issue filed, which may cause this support to be dropped later.

PR #881, opened 2026-05-09
[Flatcar] Root login enablement failed on Flatcar 4593.2.0: No python interpreters found

ZouYuhua:
GOSV-6179
Root login enablement failed on Flatcar 4593.2.0: No python interpreters found

### DIFF
--- a/linux/deploy_vm/flatcar/flatcar_post_config.yml
+++ b/linux/deploy_vm/flatcar/flatcar_post_config.yml
@@ -11,7 +11,7 @@
   ansible.builtin.debug: var=ap3_install_result
 
 - name: "Get python path in flatcar"
-  ansible.builtin.shell: which python
+  ansible.builtin.command: which python
   delegate_to: "{{ vm_guest_ip }}"
   register: python3_path_result
 
--- a/linux/deploy_vm/flatcar/flatcar_post_config.yml
+++ b/linux/deploy_vm/flatcar/flatcar_post_config.yml
@@ -10,6 +10,30 @@
 - name: "Display the result of python install"
   ansible.builtin.debug: var=ap3_install_result
 
+- name: "Get python path in flatcar"
+  ansible.builtin.shell: which python
+  delegate_to: "{{ vm_guest_ip }}"
+  register: python3_path_result
+
+- name: "Update VM's python path to in-memory inventory"
+  when:
+    - python3_path_result.stdout is defined
+    - python3_path_result.stdout | length > 0
+  block:
+    - name: "Reset vm_python for flatcar"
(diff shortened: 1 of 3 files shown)


ZouYuhua:
Thanks for Qi and Diane's analysis.

Take the solution:
-----------------
  python3 is defined in https://github.com/vmware/ansible-vsphere-gos-validation/blob/main/linux/deploy_vm/flatcar/install_python3.sh, one solution is to set vm_python for Flatcar to "/opt/bin/python" or "/opt/bin/python3" separately and keep it as "auto" for CoreOS.


Changes:
-----------------
set vm_python before run install_python3.sh


Why:set vm_python before run install_python3.sh instead of get python path after install python3.
-----------------

2026-05-11 00:53:53,011 | TASK [1_deploy_vm][Get python path in flatcar] *************
task path: /home/worker/workspace/jenkins-Ansible_Flatcar_90GA-30/ansible-vsphere-gos-validation/linux/deploy_vm/flatcar/flatcar_post_config.yml:13
fatal: [localhost -> 10.161.133.164]: FAILED! => {
    "changed": false,
    "module_stderr": "Shared connection to 10.161.133.164 closed.\r\n",
    "module_stdout": "bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8): No such file or directory\r\n/bin/sh: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8): No such file or directory\r\n/bin/sh: line 1: /usr/bin/python: No such file or directory\r\n",
    "msg": "The module failed to execute correctly, you probably need to set the interpreter.\nSee stdout/stderr for the exact error",
    "rc": 127,
    "warnings": [
        "No python interpreters found for host localhost (tried ['python3.11', 'python3.10', 'python3.9', 'python3.8', 'python3.7', 'python3.6', 'python3.5', '/usr/bin/python3', '/usr/libexec/platform-python', 'python2.7', '/usr/bin/python', 'python'])"
    ]
}
error message:
The module failed to execute correctly, you probably need to set the interpreter.
See stdout/stderr for the exact error


From the results we observed, if you want to specify the Python path directly, it must be set before running install_python3.sh. The reason is that after Python is installed in flatcar_post_config.sh, the following tasks will fail.
  name: "Get python path in flatcar"
  ansible.builtin.command: which python
  delegate_to: "{{ vm_guest_ip }}"
  register: python3_path_result

ZouYuhua:
passed the testing after code change for Flatcar & Flatcar LTS.

ZouYuhua:
@keirazhang Please help approve it if no more comment.  Thanks

keirazhang (approved):

PR #888, opened 2026-05-22
[Windows] Add taking screenshots after Windows GOSC to check desktop displayed

Tomorrow9:


windows/guest_customization/win_gosc_verify.yml
```diff
@@ -27,3 +27,20 @@
...
       include_tasks: check_timezone.yml
+    - name: "Retry to check Windows desktop displayed"
+      include_tasks:
+        file: check_desktop_display.yml
+        apply:
+          when: not (win_desktop_displayed | default(false))
+      loop: "{{ range(1, 31) }}"
+      loop_control:
+        loop_var: retry_time
+        pause: 10
```
123lzxm, line 38: Will there be too many screenshots with the pause duration = 10s? Will it be better to make it longer, e.g. 20s?

Tomorrow9: We can take a look at the testing results for a while to see if it's better to change to sleep longer or retry more times.

123lzxm (approved):

PR #889, opened 2026-05-27
Change ansible-core version in requirements.txt and update regex for checking text in screenshot

Tomorrow9:


123lzxm (approved):
LGTM, thanks.

PR #890, opened 2026-05-27
[Windows] Change to use VMware Tools install scheduled task in guest OS

Tomorrow9:


windows/wintools_complete_install_verify/install_vmtools.yml
```diff
@@ -13,68 +13,138 @@
...
-  register: wintools_install_job
+- name: "Create a scheduled task to run the VMware Tools install"
+  community.windows.win_scheduled_task:
+    name: "VMwareToolsInstall"
+    actions:
+      - path: "D:\\{{ wintools_setup_exe }}"
+        arguments: '/S /v "/qn ADDLOCAL=ALL REBOOT=R"'
+    triggers:
+      - type: time
+        start_boundary: "{{ '%Y-%m-%dT%H:%M:%S' | strftime((win_date_time | to_datetime('%Y-%m-%dT%H:%M:%S')).strftime('%s') | int + 60) }}"
```
123lzxm, line 50: It seems a little complicated. Can it be simplified?

E.g 

start_boundary: >-
  {{
    (
      win_date_time | to_datetime('%Y-%m-%dT%H:%M:%S')
      + timedelta(seconds=60)
    ).strftime('%Y-%m-%dT%H:%M:%S')
  }}

windows/wintools_complete_install_verify/install_vmtools.yml
```diff
@@ -13,68 +13,138 @@
...
+      Not get 'Running' status of service 'VMTools' in guest OS within 30s after VMware Tools install
+      scheduled task finished. Got services status dict: {{ vmtools_service_dict }}.
 
-- name: "Pause 30 seconds after VMware Tools install"
+- name: "Pause 10 seconds after VMware Tools install"
   ansible.builtin.pause:
-    seconds: 30
+    seconds: 10
+
+- name: "Clean up the scheduled task"
```
123lzxm, line 142: Does it has any impact if not cleaning up the scheduled task? If yes, then it's better to move it before the task in line 110. 
- name: "Check the status of scheduled task

windows/wintools_complete_install_verify/install_vmtools.yml
Tomorrow9: From Gemini: standard Ansible playbooks do not import the timedelta object by default into the Jinja2 rendering engine. If you try to run your snippet as-is, Ansible will likely throw an error saying datetime.timedelta or timedelta is undefined.

windows/wintools_complete_install_verify/install_vmtools.yml
Tomorrow9: If task succeeds but the scheduled task is not removed successfully in guest, I think there is no impact on the testing since it's set to be started only on the specified time.

windows/wintools_complete_install_verify/install_vmtools.yml
```diff
@@ -13,68 +13,138 @@
...
+    - name: "Get VMware Tools install scheduled task state"
+      community.windows.win_scheduled_task_stat:
+        name: "VMwareToolsInstall"
+      delegate_to: "{{ vm_guest_ip }}"
+      register: sched_task_status_2
+      until:
+        - sched_task_status_2 is defined
+        - sched_task_status_2.state.last_task_result is defined
+        - sched_task_status_2.state.last_task_result | int in [0, 3010]
+      retries: "{{ 45 | int - sched_task_status.attempts | default(0) | int }}"
```
123lzxm, line 104: If it has reached the maximum retry times for the first time(line 73), will the second retry be triggered?

If only 45 attempts are allowed, I think it's better to add the condition to "when" of the block (line 88).

windows/wintools_complete_install_verify/install_vmtools.yml
```diff
@@ -13,68 +13,139 @@
...
+    - sched_task_status.state.last_task_result | int in [0, 3010]
+  retries: 45
+  delay: 20
+  ignore_errors: true
+  ignore_unreachable: true
 
-- name: "Check VMware Tools install task finished"
+- name: "Connection to guest OS lost"
+  when: >
+    (sched_task_status.attempts is undefined or sched_task_status.attempts | int < 45) and
```
123lzxm, line 89: I'm thinking when the second retry will be trigged. 
1. If the scheduled task status was retrieved by the first check, no second retry
2. If the status was not retrieved by the first check because of the connection error, it should run out of the maximum attempts, no second retry, right?

Tomorrow9: Sometimes there are unreachable fatal errors occur, then the second retry block need to handle this situation instead of fail the test.

Tomorrow9: The second check will be triggered when 1st retries not reach the max times and connection issue happens.

123lzxm (approved):
LGTM, thanks.

PR #906, opened 2026-07-03
[Ubuntu] screenshots taken at test case failure are all black screen 

ZouYuhua:
GOSV-6363:
screenshots taken at test case failure are all black screen on Ubuntu 25.10 VM

### DIFF
--- a/common/vm_take_screenshot.yml
+++ b/common/vm_take_screenshot.yml
@@ -36,9 +36,6 @@
               - SPACE
               - SPACE
               - SPACE
-              - ENTER
-              - ENTER
-              - ENTER
               {% else %}
               - TAB
               {% endif %}
--- a/common/vm_take_screenshot.yml
+++ b/common/vm_take_screenshot.yml
@@ -28,18 +28,23 @@
     - name: "Send wake up key to active screen in case it is blank"
       when: vm_screen_active | default(false)
       block:
-        - name: "Send key to active screen (TAB for generic, CTRL_ALT_DELETE for Ubuntu)"
+        - name: "Send TAB key for Non-Ubuntu"
           include_tasks: vm_guest_send_key.yml
           vars:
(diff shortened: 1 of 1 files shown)


common/vm_take_screenshot.yml
```diff
@@ -25,14 +25,18 @@
 - name: "Take screenshot when VM is 'poweredOn'"
   when: vm_power_state_get == 'poweredOn'
   block:
-    - name: "Send a TAB key to active screen in case it is blank"
+    - name: "Send a key to active screen in case it is blank"
       include_tasks: vm_guest_send_key.yml
       vars:
         keys_send:
-          - TAB
+          - CTRL_ALT_DELETE
```
Tomorrow9: We can send this key for Ubuntu only this time and test on Ubuntu only. In the future if there is same issue found on other GOS, we can consider if apply to others then.

ZouYuhua:
@Tomorrow9 
Take vm_guest_id to check the guest OS instead of guest_os_ansible_distribution because guest_os_ansible_distribution may not defined if rerun failed testcase.

console output:
msg": "The conditional check 'guest_os_ansible_distribution != \"Ubuntu\"' failed. The error was: error while evaluating conditional (guest_os_ansible_distribution != \"Ubuntu\"): 'guest_os_ansible_distribution' is undefined. 'guest_os_ansible_distribution' is undefined\n\nThe error appears to be in '/home/worker/workspace/jenkins-Ansible_Cycle_Ubuntu_25.10_Desktop_ISO-94/ansible-vsphere-gos-validation/common/vm_take_screenshot.yml': line 28, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n  block:\n    - name: \"Send a TAB to active screen in case it is blank\"\n      ^ here\n"

PR #923, opened 2026-08-17
Update install_ovt_from_source.yml to check version to decide with "--without-x" or not

jingyanjin:
if OVT version < 13.1.0, install_ovt_from_source with option "--without-x" because GTK4 feature is only available in 13.1.0 & later

### DIFF
--- a/linux/open_vm_tools/install_ovt_from_source.yml
+++ b/linux/open_vm_tools/install_ovt_from_source.yml
@@ -26,6 +26,8 @@
   ansible.builtin.set_fact:
     ovt_dependencies: "{{ _ovt_build_config.dependencies | default([]) }}"
     ovt_configure_options: "{{ _ovt_build_config.configure_options | default([]) }}"
+    ovt_tarball_version: "{{ (linux_ovt_tarball_url | default('')) | regex_search('(\\d+\\.\\d+\\.\\d+)') }}"
+
   vars:
     _ovt_build_config: "{{ lookup('file', ovt_build_config_file_path) | from_yaml }}"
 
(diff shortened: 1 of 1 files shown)


jingyanjin:
I have validated it with 12.5.5 on RHEL/AlmaLinux/Rocky/OracleLinux 10.x.
Also run the regression with 13.2.0 on RHEL 10.x

linux/open_vm_tools/install_ovt_from_source.yml
```diff
@@ -26,6 +26,8 @@
   ansible.builtin.set_fact:
     ovt_dependencies: "{{ _ovt_build_config.dependencies | default([]) }}"
     ovt_configure_options: "{{ _ovt_build_config.configure_options | default([]) }}"
+    ovt_tarball_version: "{{ (linux_ovt_tarball_url | default('')) | regex_search('(\\d+\\.\\d+\\.\\d+)') }}"
```
Tomorrow9: Default value of `linux_ovt_tarball_url` can be omitted since there is check in `pre_ovt_install.yml`.

linux/open_vm_tools/install_ovt_from_source.yml
```diff
@@ -35,7 +37,8 @@
   when:
     - guest_os_ansible_system == 'linux'
```
Tomorrow9, line 38:
```
    - guest_os_ansible_system == 'linux'
    - (ovt_tarball_version is version('13.1.0', '<')) or
      (ovt_dependencies | select('search', 'gtk[^0-9]*4') | length < 2)
```

Tomorrow9: Is this the expected condition?

jingyanjin: Yes

PR #926, opened 2026-08-18
Retain the original description if a new one is not provided when renaming a snapshot

123lzxm:


Tomorrow9:
Can this merge request be closed since there is fix in Ansible module? Thanks.

PR #927, opened 2026-08-21
Change to create VMkernel network adapter without specifying pre-generated name

Tomorrow9:
To avoid race condition when several jobs are running in this network environment setup tasks at the same time in parallel on the same testbed.


linhuali (approved):
LGTM. Thanks.

